PR #30, opened 2021-12-30
Add coverage data, fix docker multistage

tonvanbart:
- added jacoco plugin to Maven build
- run workflow on branches recognized as long lived by Sonarcloud (`branch-*`)
- need to add a workflow secret for SONAR_TOKEN

Verified on my fork that coverage is now shown in Sonarcloud.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -91,7 +91,7 @@
         <license.licensename>apache_v2</license.licensename>
 
         <!-- sonarcloud properties -->
-        <sonar.organization>tonvanbart</sonar.organization>
+        <sonar.organization>axual</sonar.organization>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -91,8 +91,12 @@
         <license.licensename>apache_v2</license.licensename>
 
         <!-- sonarcloud properties -->
-        <sonar.organization>axual</sonar.organization>
+        <sonar.organization>tonvanbart</sonar.organization>
```
jeroenvandisseldorp, line 94: Should we rename this back to axual before merging?

tonvanbart: Yep

PR #50, opened 2022-06-07
Support Liberica NIK

JRial95:
Currently, it's not possible to use Liberica NIK to compile and test because the `checkGraalVM()` method fails : 

```
org.opentest4j.AssertionFailedError: This test needs GraalVM to work, found java.vendor.url=https://bell-sw.com/
	at io.axual.ksml.KSMLTopologyGeneratorBasicTest.checkGraalVM(KSMLTopologyGeneratorBasicTest.java:57)
```

As mentioned by Oleg Šelajev in this [StackOverflow thread](https://stackoverflow.com/questions/64902270/how-can-i-tell-in-a-standard-java-program-if-im-running-in-graalvm-or-not), a cleaner (and generic) way to check if we use GraalVM is to use the [`Version.isRelease()`](https://www.graalvm.org/sdk/javadoc/org/graalvm/home/Version.html) method from the `graal-sdk`.


tonvanbart (approved):
Nice one, thanks!

Issue #58, opened 2022-11-15
Base image should be updated to 8.6 release


richard-axual:
The base image, Red Hat UBI 8.4, needs to be updated to the newer 8.6 version.
Latest tag at this time is 8.6-990

richard-axual:
Merged to master

PR #59, opened 2022-11-15
Update dockerfile to use newer UBI base image

richard-axual:
Update the docker base image to UBI 8.6-990

Fix for issue #58


tonvanbart (approved):
LGTM

PR #63, opened 2022-11-23
Implement schema support

jeroenvandisseldorp:
Adds basic schema support, plus fixes many smaller issues.


examples/test.avsc
```diff
@@ -0,0 +1,68 @@
+{
```
richard-axual, line 1: This file is named test.
If this is part of the examples it should be renamed. Or else it should be removed

richard-axual (changes requested):

tonvanbart:
There are 242 changed files in this PR, which makes it a bit unwieldy to review. Some (most?) of the changes appear to be non related to schema support, or is that just me?
It would be better if we split this PR into two, one for the formatting/small issues changes and one for the new functionality.

ksml/src/main/java/io/axual/ksml/data/mapper/UserTypeStringMapper.java
```diff
@@ -0,0 +1,102 @@
...
+        if (value instanceof ListType listType) return listTypeToString(listType);
+        if (value instanceof StructType structType) return structTypeToString(structType);
+        if (value instanceof TupleType tupleType) return tupleTypeToString(tupleType);
+        return null;
+    }
+
+    private String enumTypeToString(EnumType enumType) {
+        var result = new StringBuilder("enum(");
+        var first = true;
+        for (String symbol : enumType.symbols()) {
```
tonvanbart, line 71: Could be made simpler using streams and Collectors.joining()

ksml/src/main/java/io/axual/ksml/data/mapper/UserTypeStringMapper.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public String toString(UserType value) {
+        var notation = value.notation().toLowerCase();
+        var type = toString(value.dataType());
+        if (type == null || type.isEmpty()) return notation;
+        return notation + ":" + type;
+    }
+
+    private String toString(DataType value) {
+        if (value == DataNull.DATATYPE) return "null";
```
tonvanbart, line 51: Might be better to move this into some toString() method inside the enum

tonvanbart (changes requested):
None of the added files have any test coverage, I think there should be some test code at least.

examples/test.avsc
jeroenvandisseldorp: Removed

examples/test1.json
```diff
@@ -0,0 +1,7 @@
+{"schema":"{"type":"record","name":"SensorData","namespace":"io.axual.ksml.example","doc":"","fields":[{"name":"name","type":"string","doc":""},{"name":"timestamp","type":"long","doc":""},{"name":"value","type":"string","doc":""},{"name":"type","type":{"type":"enum","name":"SensorType","doc":"","symbols":["AREA","HUMIDITY","LENGTH","STATE","TEMPERATURE"]},"doc":""},{"name":"unit","type":"string","doc":""},{"name":"color","type":["null","string"],"doc":"","default":null},{"name":"city","type":["null","string"],"doc":"","default":null},{"name":"owner","type":["null","string"],"doc":"","default":null}]}"}
```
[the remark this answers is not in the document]
jeroenvandisseldorp: Removed

ksml/src/main/java/io/axual/ksml/data/mapper/UserTypeStringMapper.java
jeroenvandisseldorp: I considered this, but the mapper goes two-way, namely from/to the formal syntax of the UserType. Implementing this in the class itself would mean that the class itself knows its formal syntax. Normally toString() in a class is only used to aid developers or for one-way conversion to String, but not to be parsed back.

jeroenvandisseldorp:
> There are 242 changed files in this PR, which makes it a bit unwieldy to review. Some (most?) of the changes appear to be non related to schema support, or is that just me? It would be better if we split this PR into two, one for the formatting/small issues changes and one for the new functionality.

Valid point, and discussed with Richard as well last week, but not something we can do much about at this stage.

ksml/src/main/java/io/axual/ksml/data/mapper/UserTypeStringMapper.java
jeroenvandisseldorp: Done

richard-axual (approved):
Approved, but we really need to start working on tests and refactoring to make KSML easy to test

PR #64, opened 2022-11-25
Cleaned up code smells

jeroenvandisseldorp:
Code cleanups after Sonar check


ksml-runner/src/test/java/io/axual/ksml/runner/config/KSMLRunnerConfigTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class KSMLRunnerConfigTest {
```
tonvanbart, line 35: This class seems to be duplicate and fails the pipeline, should it be moved or renamed?

tonvanbart (changes requested):

ksml-runner/src/test/java/io/axual/ksml/runner/config/KSMLRunnerConfigTest.java
jeroenvandisseldorp: Renamed the class to have uppercase KSML instead of Camel case. But for some reason the filename was not renamed in the same way. Corrected.

tonvanbart (approved):

PR #81, opened 2023-11-03
Manual state store support, Kafka client cleanups and configuration changes

jeroenvandisseldorp:


ksml-kafka-clients/pom.xml
```diff
@@ -10,35 +10,39 @@
         <version>0.1.1-SNAPSHOT</version>
     </parent>
 
-    <artifactId>ksml-axual</artifactId>
-    <name>KSML for Axual</name>
-    <description>Additional components for KSML to enable execution with Axual servuces
+    <artifactId>ksml-kafka-clients</artifactId>
+    <name>Kafka clients for KSML</name>
+    <description>Client components for KSML
```
richard-axual, line 15: Maybe put a bit more information here what functionality these components offer?

ksml-kafka-clients/src/main/java/io/axual/ksml/client/util/FactoryUtil.java
richard-axual, line 1: Why rebuild this logic here instead of using the Kafka Utils directly?.

ksml-kafka-clients/src/main/java/io/axual/ksml/client/admin/ProxyAdmin.java
richard-axual, line 1: I think this can be removed, or replaced by the normal Kafka Admin interface, as this proxy is initialised explicitly with an already build Admin instance

ksml-kafka-clients/src/main/java/io/axual/ksml/client/consumer/ProxyConsumer.java
richard-axual, line 1: I think this can be removed, or replaced by the normal Kafka Consumer interface, as this proxy is initialised explicitly with an already build Consumer instance

ksml-kafka-clients/src/main/java/io/axual/ksml/client/producer/ProxyProducer.java
richard-axual, line 1: I think this can be removed, or replaced by the normal Kafka Admin interface, as this proxy is initialised explicitly with an already build Admin instance

richard-axual (changes requested):
It looks good overall, but I think several of the proxy classes are no longer needed.

ksml-data-generator/src/main/java/io/axual/ksml/datagenerator/config/ProducerConfig.java
```diff
@@ -20,9 +20,12 @@
...
  */
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import java.util.List;
 
-public class DataGeneratorConfig {
+public class ProducerConfig {
+    @JsonProperty("working.directory")
     public String workingDirectory;
```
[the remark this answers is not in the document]
jeroenvandisseldorp: I hadn't noticed before, but there is actually a difference now between working directory and configuration directory. The working directory holds ksml-data-generator.yaml and ksml-runner.yaml, and it acts as the folder for local state store storage. The config directory holds all all .yaml files to be read (except for the ones mentioned above).

We could do like this:
- Configuration dir: contains all .yaml files, schema etcetera. Also make this the default location for ksml-runner.yaml and ksml-data-generator.yaml too. Allow to be overriden by first parameter to the executable.
- Working dir: rename to "storage dir" --> only required for runner, not for data generator

ksml-data-generator/src/main/java/io/axual/ksml/datagenerator/factory/AxualClientFactory.txt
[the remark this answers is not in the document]
jeroenvandisseldorp: Yes, done

ksml-kafka-clients/src/main/java/io/axual/ksml/client/admin/ProxyAdmin.java
jeroenvandisseldorp: The idea is that ProxyXX classes call the backing client by default. The ResolvingXX classes only override those methods that actually need to do resolving. Makes for better readability, that's all. If this is deemed over-engineered, then we could remove them.

ksml-kafka-clients/src/main/java/io/axual/ksml/client/consumer/ProxyConsumer.java
jeroenvandisseldorp: See ProxyAdmin comment

ksml-data-generator/src/main/java/io/axual/ksml/datagenerator/config/ProducerConfig.java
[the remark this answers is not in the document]
richard-axual: @jeroenvandisseldorp I agree with your suggestion to make config dir the container of YAMLs, Schemas etc and rename working dir to storage dir

ksml-kafka-clients/src/main/java/io/axual/ksml/client/admin/ProxyAdmin.java
richard-axual: @jeroenvandisseldorp I think inheritance and delegate methods are overkill for this, we don't expect any other classes needing to inherit from.

ksml-data-generator/src/main/java/io/axual/ksml/datagenerator/config/ProducerConfig.java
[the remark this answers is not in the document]
richard-axual: @jeroenvandisseldorp The runner configuration has been updated to the new directory structure, can you update the DataGenerator config as well?

richard-axual (approved):

PR #83, opened 2023-11-27
Update documentation to use new runner configurations

richard-axual:
Update documentation for Axual Runner
Fix final examples


richard-axual:
Merged because the missing documentation was needed

PR #86, opened 2023-12-07
Rework parsing and topology generation logic, together with DSL improvements

jeroenvandisseldorp:
This update redoes the parsing logic. Parsing and topology generation are now separate steps:

1. Parse a definition into internal Definition structures
2. Parse producer and/or pipeline definitions, using the aforementioned Definitions
3. Generate the producer and/or pipeline logic.
4. Run the logic

In this PR there is a lot of unification of nomenclature and configuration options, eg. for Materialized, Joined, StreamJoined configuration options. Also the DSL is cleaned up, providing better insight into the YAML substructures per definition element.


richard-axual (approved):
A lot of work, but its good

PR #97, opened 2024-04-05
Make `producers:` interval optional so we can single shot produce

tonvanbart:
While working on / thinking about testing KSML streams I noticed that the `interval:` parameter inside `producers:` was mandatory. If we want to use KSML to create tests for stream definitions I think it would be useful to be able to create single-shot producers which generate a known, fixed set of records on a stream.
This PR makes `interval:` optional; if omitted, the produce will run one time only.

Added a unit test for `IntervalSchedule` to verify behavior, tested in docker compose by changing `00-example-generate-sensordata.yaml` to have no interval, running with a local generated image and verified that only one sensordata reading was generated.

### DIFF
--- a/ksml/src/test/java/io/axual/ksml/TopologyGeneratorBasicTest.java
+++ b/ksml/src/test/java/io/axual/ksml/TopologyGeneratorBasicTest.java
@@ -91,6 +91,19 @@ public class TopologyGeneratorBasicTest {
         assertThat(cleanDescription("fix\r\nnewlines"), is("fix\nnewlines"));
     }
 
+    @Test
+    void loadGeneratorExample() throws Exception {
+        final var uri = ClassLoader.getSystemResource("pipelines/generator-example.yaml").toURI();
+        final var path = Paths.get(uri);
+        final var definition = YAMLObjectMapper.INSTANCE.readValue(Files.readString(path), JsonNode.class);
+        final var definitions = ImmutableMap.of("definition",
+                new TopologyDefinitionParser("test").parse(ParseNode.fromRoot(definition, "test")));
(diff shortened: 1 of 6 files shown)


ksml/src/test/java/io/axual/ksml/TopologyGeneratorBasicTest.java
```diff
@@ -91,6 +91,19 @@ void cleanDescriptionTestLineEnd() {
         assertThat(cleanDescription("fix\r\nnewlines"), is("fix\nnewlines"));
     }
 
+    @Test
+    void loadGeneratorExample() throws Exception {
```
richard-axual, line 95: Can you clarify what is tested here?
As I see it it tests that the file can be successfully parsed to a topology?
Can we rename the test to something more descriptive, maybe without the example in the name?

richard-axual (changes requested):
This looks good.
Just one remark about the new test in TopologyGeneratorBasicTest.

ksml/src/test/java/io/axual/ksml/TopologyGeneratorBasicTest.java
tonvanbart: Hi Richard, the intention was to have a test proving that `interval:` can be optional, this is now in `IntervalScheduleTest`. Removed the test and the resource; also made the test conditional on GraalVM (`@EnabledIf`) instead of the ugly fail if not Graal.

tonvanbart:
At the request of @jeroenvandisseldorp , also added parsing and handling for `count` (fixed number of times) and `until` (until a predicate becomes true). These can be combined.

richard-axual (approved):

PR #98, opened 2024-04-09
Enabled Topology Generator Basic test to verify topology and topology naming

richard-axual:
During the refactoring the topology generator tests were partially disabled and never reactivated.

Updated the tests to match the current initialisation logic.
Updated the reference files to match the expected topology names


tonvanbart (approved):
LGTM

PR #144, opened 2024-07-17
Add proper stream data type flattening for transform operations

jeroenvandisseldorp:


richard-axual (approved):
Tested with previously failing examples and with the provided project examples. No regressions found

PR #165, opened 2025-01-13
Fix store serde regression

jeroenvandisseldorp:
Revert back to default store serde in all cases. This patch fixes a wrong assumption on compaction topics, where an internal serde was used instead of the default serde provided by the backing topic.


richard-axual (approved):
Build and run on my local machine, no issues found

PR #166, opened 2025-01-14
Update several dependencies

jeroenvandisseldorp:


richard-axual (changes requested):
Looks good, but can you add the maven versions plugin 2.18.0 to the parent pom pluginManagement?

This allows us to easily update versions using maven
```xml
                <plugin>
                    <groupId>org.codehaus.mojo</groupId>
                    <artifactId>versions-maven-plugin</artifactId>
                    <version>${maven.versions.plugin.version}</version>
                    <configuration>
                        <allowMajorUpdates>false</allowMajorUpdates>
                    </configuration>
                </plugin>
```

example update command is 
```shell
mvn versions:update-properties
mvn versions:commit
mvn versions:revert
```

PR #167, opened 2025-01-24
Introduce protobuf

jeroenvandisseldorp:
This PR introduces Protobuf encoding/decoding to KSML. It does so through Apicurio serdes, which can be configured from the main ksml-runner.yaml. Other schema-based serdes (AVRO only atm), like Apicurio's AVRO and Confluent's AVRO, can be configured in the same YAML as the default for their respective schema type.


richard-axual (changes requested):
@jeroenvandisseldorp The current code is missing tests for most changes and additions.
I think we need to have at least serialisation/deserialisation tests, to make sure the conversion to and from protobuf is working as expected.

Can you add those changes?

jeroenvandisseldorp:
> @jeroenvandisseldorp The current code is missing tests for most changes and additions. I think we need to have at least serialisation/deserialisation tests, to make sure the conversion to and from protobuf is working as expected.
> 
> Can you add those changes?

Added

richard-axual (approved):

PR #172, opened 2025-01-29
Fix several storage issues and allow storageDIrectory creation

richard-axual:
There were several issues related to the state storage.
The first was that the Kubernetes mounted volume did not match the ownership as the default user in the KSML image.
This was fixed by creating the KSML user with primary group 0 (root)

There is also a change that will allow KSML to create the storage directory if it does not exist.
A new  config parameter has been added to enable creation of the storage directory

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,7 +72,7 @@ ENV JAVA_HOME=/opt/graalvm
 COPY --chown=ksml:0 --from=graal-builder /opt/graal/ /opt/graal/
 
 WORKDIR /home/ksml
-USER 1024:0
+USER 1024
 #There is no more GraalPy command here and no venv
 #RUN graalpy -m venv graalenv && \
 #    echo "source $HOME/graalenv/bin/activate" >> ~/.bashrc
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,7 @@ ENV PATH=/opt/graal/bin:$PATH \
 RUN set -eux \
     	&& microdnf upgrade -y --nodocs \
         && microdnf install -y --nodocs procps tar unzip gzip zlib openssl-devel gcc gcc-c++ make patch glibc-langpack-en libxcrypt shadow-utils \
-	&& groupadd --gid 1024 ksml \
-        && useradd -g ksml -u 1024 -d "$KSML_HOME" -ms /bin/sh -f -1 ksml \
+        && useradd -g 0 -u 1024 -d "$KSML_HOME" -ms /bin/sh -f -1 ksml \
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -73,7 +72,7 @@ ENV JAVA_HOME=/opt/graalvm
 COPY --chown=ksml:0 --from=graal-builder /opt/graal/ /opt/graal/
 
 WORKDIR /home/ksml
-USER 1024
+USER 1024:0
```
xsreality, line 75: This is not required as the user is created in root group already.

examples/ksml-runner.yaml
```diff
@@ -3,6 +3,7 @@ ksml:
   configDirectory: .              # When not set defaults to the working directory
   schemaDirectory: .              # When not set defaults to the config directory
   storageDirectory: /tmp          # When not set defaults to the default JVM temp directory
+  createStorageDirectory: false   # When set to true, the storage directory will be created if it does not exist
```
xsreality, line 6: Shall we keep the default to true since almost in all cases it will have to be enabled to avoid errors?

ksml-runner/src/main/java/io/axual/ksml/runner/config/KSMLConfig.java
```diff
@@ -58,9 +60,14 @@ public class KSMLConfig {
     @Builder.Default
     @Getter
     private PrometheusConfig prometheusConfig = DEFAULT_PROMETHEUS_CONFIG;
+
     private String configDirectory;
     private String schemaDirectory;
     private String storageDirectory;
```
xsreality, line 66: The `storageDirectory` config is [hard-coded in the helm charts](https://github.com/Axual/ksml/blob/main/packaging/helm-charts/ksml/templates/configmap-runner.yaml#L12). This needs to be made configurable via values.yaml.

xsreality (changes requested):

examples/ksml-runner.yaml
richard-axual: I've made it false to not break the existing plain KSML images running, but enabled it by default in the Helm chart

ksml-runner/src/main/java/io/axual/ksml/runner/config/KSMLConfig.java
richard-axual: It's configurable now, and the charts now allow additional claim templates that your can use

Dockerfile
richard-axual: Removed the explicit group id

xsreality (approved):

PR #205, opened 2025-05-04
Make Graal‑VM security options configurable via YAML

KasparMetsa:
### 🚀  What’s new
* **PythonContextConfig**
  * Added 6 new boolean flags with sensible defaults (`false`):
    * `allowHostFileAccess`
    * `allowHostSocketAccess`
    * `allowNativeAccess`
    * `allowCreateProcess`
    * `allowCreateThread`
    * `inheritEnvironmentVariables`
* **PythonContext**
  * Replaced these 6 hard‑coded settings with values taken from `PythonContextConfig`.
  * We map `inheritEnvironmentVariables`
    * `true`  ➜ `EnvironmentAccess.INHERIT`
    * `false` ➜ `EnvironmentAccess.NONE`
  * Extended debug log to print the chosen options.

### YAML usage
```yaml
ksml:
  pythonContext:
    allowHostFileAccess:         false
    allowHostSocketAccess:       false
    allowNativeAccess:           false
    allowCreateProcess:          false
    allowCreateThread:           false
    inheritEnvironmentVariables: false

### CODE REVIEW

File: ksml/src/main/java/io/axual/ksml/python/PythonContext.java [hunk 1]
```diff
@@ -43,19 +43,28 @@ public class PythonContext {
     @Getter
     private final DataObjectConverter converter;
 
-    public PythonContext() {
+    public PythonContext(PythonContextConfig config) {
         this.converter = new DataObjectConverter();
 
-        log.debug("Setting up new Python context");
+        log.debug("Setting up new Python context (fileAccess={}, socketAccess={}, nativeAccess={}, " +
```
richard-axual, line 49: If you create a toString method in the PythonContextConfig class you can simplify this.
You can use to lombok ToString annotation and specify which fields to use

ksml/src/test/java/io/axual/ksml/python/PythonContextConfigTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+class PythonContextConfigTest {
+
+    private final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+    @Test
+    @DisplayName("Builder defaults all flags to false")
+    void defaultBuilderFlags() {
+        var cfg = PythonContextConfig.builder().build();
+        assertFalse(cfg.allowHostFileAccess(),     "allowHostFileAccess should default to false");
```
richard-axual, line 18: We also have assertJ available.
Using SoftAssertions you can use a more readable assertion, and actually test all values at once.

richard-axual (changes requested):
We don't have a PythonContext test at this time.
Can you add it and check if the configurable restrictions are working

verifyHostFileDenied, verifyHostFileAllowed, etc

With just a simple Python script that tries to do a single call, like read file

richard-axual (approved):

PR #206, opened 2025-05-06
Implement Job deployment mode for data generator

richard-axual:
The standard deployment for KSML is as a Kubernetes StatefulSet. 
This provides horizontal scaling options with predictable naming and volume claims, which is great for pipeline processing.
The Kubernetes cluster will restart completed applications to meet the replica requirement.

But this is unwanted if KSML is configured as a data generator with a fixed number of produce calls, or other ending condition.
By using a Kubernetes Job the data generation will run only once, and keep the pod in its final state to allow deployment and log inspection.

This PR adds the `deploymentMode` configuration option to deploy the application as either a StatefulSet or a Job.

Known limitations:
* Job deployments can not be updated, but must be removed and recreated. This is a standard Kubernetes Job limitation
* Jobs can not scale horizontally


packaging/helm-charts/ksml/templates/Job.yaml
```diff
@@ -0,0 +1,101 @@
...
+kind: Job
+metadata:
+  name: {{ include "ksml.fullname" . }}
+  labels:
+    {{- include "ksml.labels" . | nindent 4 }}
+spec:
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap-runner.yaml") . | sha256sum }}
```
KasparMetsa, line 12: Can we consider adding a checksum annotation for logging config(ksml-logging ConfigMap) to trigger Pod restarts on changes:

`checksum/logging: {{ include (print $.Template.BasePath "/configmap-logging.yaml") . | sha256sum }}
`

packaging/helm-charts/ksml/templates/Job.yaml
```diff
@@ -0,0 +1,101 @@
+{{- if eq "job" (lower .Values.deploymentMode) }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "ksml.fullname" . }}
+  labels:
+    {{- include "ksml.labels" . | nindent 4 }}
+spec:
+  template:
```
KasparMetsa, line 9: Can we consider adding:

1.  `ttlSecondsAfterFinished: 3600` to clean up old Jobs automatically. If unset, Jobs and their Pods are retained forever until manually deleted.
2. `backoffLimit: 0` to disable retries if pod fails. K8s default is 6 retries.

packaging/helm-charts/ksml/templates/Job.yaml
```diff
@@ -0,0 +1,101 @@
...
+      volumes:
+        - name: ksml-config
+          configMap:
+            name: {{ include "ksml.fullname" . }}-runner
+        - name: ksml-logging
+          configMap:
+            name: {{ include "ksml.fullname" . }}-logging
+        - name: ksml-store
+          emptyDir:
+            sizeLimit: 10Mi
```
KasparMetsa, line 76: Maybe good to make this configurable?

KasparMetsa (changes requested):

packaging/helm-charts/ksml/templates/Job.yaml
richard-axual: Restarts are not needed, it's Logback based and the configmap is configured to reload at runtime

packaging/helm-charts/ksml/templates/Job.yaml
richard-axual: Good point, but I'll set the ttlSecondsAfterFinished default to 259200 (3 days) to allow inspection, even over the weekend. Our use case usually don't have short lived jobs.
Now that I think of this, maybe add a specific configuration for optional finalizers as well, to control cleanup.

packaging/helm-charts/ksml/templates/Job.yaml
richard-axual: Maybe a future improvement, we would also need this for the StatefulSet

KasparMetsa (approved):

PR #207, opened 2025-05-08
refactor: remove unused "index" field in MemberType

tonvanbart:
Analyzing the code revealed that `UnionType.MemberType` had a field `index` that was never used; array ordering in the `UnionType` is used for member ordering.
Refactored the code to remove `index` and verified that all tests still work.


jeroenvandisseldorp (commented):
The index field is not an unused field. It is used during conversion to Protobuf schema output, where each union member type has a tag (or index) associated with it.

tonvanbart:
> The index field is not an unused field. It is used during conversion to Protobuf schema output, where each union member type has a tag (or index) associated with it.

OK - I checked (as an experiment) with the Jetbrains AI agent which indicates that `index` was never used in serialization but that it was all based on array indices, that's why the change seemed safe (and also because the build did not break).

What would we need to add as an integration test to verify that serialization still works for all schema types after a software change? Maybe some test where we copy for ex. an AVRO stream1 to Protobuf stream2 and then back to AVRO stream3, then comparing the contents of stream1 and stream3? Could that work?

tonvanbart:
Closed, as the PR was based on an incorrect assumption.

PR #223, opened 2025-06-10
Fix regression because of a SOAP XML dependency

jeroenvandisseldorp:


ksml-data-soap/pom.xml
```diff
@@ -28,6 +29,16 @@
             <groupId>jakarta.xml.soap</groupId>
             <artifactId>jakarta.xml.soap-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.sun.xml.messaging.saaj</groupId>
+            <artifactId>saaj-impl</artifactId>
```
richard-axual, line 34: Can you put this to scope runtime if its not needed for compilation?

richard-axual (approved):

PR #231, opened 2025-06-12
Improve the window handling of joins with streams

jeroenvandisseldorp:
* Fix in WindowStateStoreDefinition that copied the wrong value into its windowSize
* Checks in BaseOperation to validate window store parameters with JoinWindows
* Update in TopologyGenerator to skip manual creation of join operation state stores

### DIFF
--- a/ksml/src/main/java/io/axual/ksml/definition/WindowStateStoreDefinition.java
+++ b/ksml/src/main/java/io/axual/ksml/definition/WindowStateStoreDefinition.java
@@ -37,7 +37,7 @@ public final class WindowStateStoreDefinition extends StateStoreDefinition {
     public WindowStateStoreDefinition(String name, Boolean persistent, Boolean timestamped, Duration retention, Duration windowSize, Boolean retainDuplicates, UserType keyType, UserType valueType, Boolean caching, Boolean logging) {
         super(StoreType.WINDOW_STORE, name, persistent, timestamped, keyType, valueType, caching, logging);
         this.retention = retention != null ? retention : Duration.ZERO;
-        this.windowSize = windowSize != null ? retention : Duration.ZERO;
+        this.windowSize = windowSize != null ? windowSize : Duration.ZERO;
         this.retainDuplicates = retainDuplicates != null && retainDuplicates;
     }
 }
--- a/ksml/src/main/java/io/axual/ksml/operation/BaseOperation.java
+++ b/ksml/src/main/java/io/axual/ksml/operation/BaseOperation.java
@@ -42,6 +42,7 @@ import org.apache.kafka.streams.kstream.*;
 import org.apache.kafka.streams.processor.StreamPartitioner;
 import org.apache.kafka.streams.state.KeyValueStore;
 import org.apache.kafka.streams.state.SessionStore;
+import org.apache.kafka.streams.state.WindowBytesStoreSupplier;
 import org.apache.kafka.streams.state.WindowStore;
 
 import java.time.Duration;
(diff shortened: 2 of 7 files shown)


ksml/src/main/java/io/axual/ksml/definition/WindowStateStoreDefinition.java
```diff
@@ -37,7 +37,7 @@ public final class WindowStateStoreDefinition extends StateStoreDefinition {
     public WindowStateStoreDefinition(String name, Boolean persistent, Boolean timestamped, Duration retention, Duration windowSize, Boolean retainDuplicates, UserType keyType, UserType valueType, Boolean caching, Boolean logging) {
         super(StoreType.WINDOW_STORE, name, persistent, timestamped, keyType, valueType, caching, logging);
         this.retention = retention != null ? retention : Duration.ZERO;
-        this.windowSize = windowSize != null ? retention : Duration.ZERO;
+        this.windowSize = windowSize != null ? windowSize : Duration.ZERO;
         this.retainDuplicates = retainDuplicates != null && retainDuplicates;
     }
```
KasparMetsa, line 42: Can we consider adding a check like this in the constructor. Checks that retention period must be at least as long as the window size.

```java
    if (this.retention.compareTo(this.windowSize) < 0) {
        throw new IllegalArgumentException("Retention period must be >= window size");
    }
```

ksml/src/main/java/io/axual/ksml/operation/BaseOperation.java
```diff
@@ -379,16 +380,31 @@ protected Joined<Object, Object, Object> joinedOf(StreamDataType k, StreamDataTy
...
         }
         if (otherStore != null) {
-            result = result.withOtherStoreSupplier(StoreUtil.getStoreSupplier(otherStore));
+            result = result.withOtherStoreSupplier(validateWindowStore(otherStore, joinWindows));
+        }
+        return result;
+    }
+
+    private WindowBytesStoreSupplier validateWindowStore(WindowStateStoreDefinition store, JoinWindows joinWindows) {
+        // Copied these validation rules from Kafka Streams' KStreamImplJoin::assertWindowSettings
```
KasparMetsa, line 398: Currently, each validation throws its own TopologyException immediately when it fails. Have we considered collecting all violations first and throwing a single exception listing all validation errors at once?

ksml/src/main/java/io/axual/ksml/definition/WindowStateStoreDefinition.java
```diff
@@ -37,7 +37,7 @@ public final class WindowStateStoreDefinition extends StateStoreDefinition {
     public WindowStateStoreDefinition(String name, Boolean persistent, Boolean timestamped, Duration retention, Duration windowSize, Boolean retainDuplicates, UserType keyType, UserType valueType, Boolean caching, Boolean logging) {
         super(StoreType.WINDOW_STORE, name, persistent, timestamped, keyType, valueType, caching, logging);
```
KasparMetsa, line 38: Would checks like these be useful here?

```java
        Objects.requireNonNull(name, "Window store name cannot be null");
        Objects.requireNonNull(keyType, "Key type cannot be null");
        Objects.requireNonNull(valueType, "Value type cannot be null");
```

ksml/src/main/java/io/axual/ksml/definition/WindowStateStoreDefinition.java
jeroenvandisseldorp: I would refrain from duplicating the validations that Kafka Streams does already. The only reason for adding a validation is because the Kafka Streams error may not provide our users the proper feedback they need. Hence the additional (and overlapping) validation done in BaseOperation.

ksml/src/main/java/io/axual/ksml/operation/BaseOperation.java
jeroenvandisseldorp: Not ATM, but worth considering. Not in this PR I think though.

ksml/src/main/java/io/axual/ksml/definition/WindowStateStoreDefinition.java
jeroenvandisseldorp: Yes, but only for type and name. Other fields are often optional in the context in which they're used. Added to the PR.

PR #241, opened 2025-06-19
Fix type conversion

jeroenvandisseldorp:
* Add better failure handling to data type conversion
* Introduce schemaless StructSchema
* Make StructType aware of schemaless type


tonvanbart:
Are we at a point that we could add a `@KSMLTest` that would fail if this bug exists? (It would prove that the bug is fixed, and protect against accidental regressions)

KasparMetsa (approved):

PR #245, opened 2025-06-23
Add support for templated KSML tests

tonvanbart:
Since some KSML operations have multiple aliases and/or can be configured in different ways, the current tests have some duplication where the same code is just annotated with a different pipeline definition.

This PR adds a new annotation `@KSMLTopologyTest` which can be provided with a list of KSML topologies; it will then run the test for each topology. `@KSMLTopic` and `@KSMLDriver` annotations are supported as before, as well as the `schemaDirectory` parameter for AVRO support.
The test class no longer needs an `@ExtendsWith` anntation, this is handled with the code that manages the new annotation.

Usage example: 
```java
    @KSMLTopologyTest(
            topologies = {
                    "pipelines/test-mapvalue-expression.yaml",
                    "pipelines/test-mapvalues-expression.yaml",
                    "pipelines/test-transformvalue-expression.yaml",
                    "pipelines/test-mapvalue-code.yaml",
                    "pipelines/test-mapvalues-code.yaml",
                    "pipelines/test-transformvalue-code.yaml"},
            schemaDirectory = "schemas")
    @DisplayName("Values can be mapped with mapValue and aliases")
    void testMapValueByExpression() {

```

### DIFF
--- a/ksml/src/test/java/io/axual/ksml/data/mapper/DataFlattenerTest.java
+++ b/ksml/src/test/java/io/axual/ksml/data/mapper/DataFlattenerTest.java
@@ -1,5 +1,25 @@
 package io.axual.ksml.data.mapper;
 
+/*-
+ * ========================LICENSE_START=================================
+ * KSML
+ * %%
+ * Copyright (C) 2021 - 2025 Axual B.V.
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


richard-axual (approved):

richard-axual:
Updated the implementation to use implicit types, and use of final variables where possible

jeroenvandisseldorp (approved):

PR #253, opened 2025-07-13
Improvements inspired by the documentation rewrite

jeroenvandisseldorp:
* Introduction of jsonschema notation
* Explicit Schema Registry configuration in ksml-runner.yaml
* Notation configuration updates in ksml-runner.yaml
* Inclusion of Confluent serdes for JSON Schema and Protobuf
* Introduction of a TupleSchema, which gets dynamically created based on a specific tuple type
* Introduction of name, version and description fields to allow metadata capturing in KSML topology definitions
* Make function expressions optional and multiline, allowing for more flexibility to return values from function definitions
* Introduce an optional partitioner field for streams, tables and globalTables, allowing for custom logic to produce messages to specific topic partitions
* Upgrade the StreamPartitioner function type to the new Java signature, allowing it to return a list of partition numbers instead of just one
* Introduce a BaseTableDefinitionParser for less code duplication between table and globalTable parsing
* Allow offset reset policy to use KSML-style duration definitions
* Lots of small corrections in descriptions and comments in source code files
* Incremental dependency updates


ksml/src/main/java/io/axual/ksml/definition/FunctionDefinition.java
```diff
@@ -77,21 +85,23 @@ public FunctionDefinition withoutResult() {
     }
 
     public FunctionDefinition withResult(UserType resultType) {
-        return new FunctionDefinition(type, name, parameters, globalCode, code, resultType != null ? expression : null, resultType, storeNames);
+        return new FunctionDefinition(type, name, parameters, globalCode, code, resultType != null ? expression : EMPTY_STRING_ARRAY, resultType, storeNames);
```
KasparMetsa, line 88: How about we validate and guide the user if expression is null or empty?

```
public FunctionDefinition withResult(UserType resultType) {
    if (resultType != null && (expression == null || expression.length == 0)) {
        throw functionResultError("Function with result type requires at least one expression line");
    }
    return new FunctionDefinition(type, name, parameters, globalCode, code, 
        resultType != null ? expression : EMPTY_STRING_ARRAY, resultType, storeNames);
}
```

ksml-data/src/main/java/io/axual/ksml/data/schema/TupleSchema.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class TupleSchema extends StructSchema {
+    public TupleSchema(TupleType type, DataTypeDataSchemaMapper mapper) {
+        super(DataSchemaConstants.DATA_SCHEMA_KSML_NAMESPACE, type.toString(), "Tuple with " + type.subTypeCount() + " fields", convertTupleTypeToFields(type, mapper));
+    }
+
+    private static List<DataField> convertTupleTypeToFields(TupleType type, DataTypeDataSchemaMapper mapper) {
+        final var result = new ArrayList<DataField>();
```
KasparMetsa, line 35: Would this validation be useful here?

```
if (type.subTypeCount() == 0) {
        throw new SchemaException("TupleSchema requires at least one field");
    }
```

ksml/src/main/java/io/axual/ksml/parser/DurationParser.java
```diff
@@ -77,6 +77,7 @@ private Duration parseDuration(String durationStr) {
             // If the duration does not contain a valid unit string, assume it is a whole number in millis
             return Duration.ofMillis(Long.parseLong(durationStr));
         } catch (NumberFormatException e) {
+            if (allowFail) return null;
             throw new TopologyException("Illegal duration: " + durationStr);
```
KasparMetsa, line 81: Can we consider improving this error message:
```
        throw new TopologyException(String.format(
            "Invalid duration format: '%s'. Expected format: <number><unit> " +
            "where unit is one of: ms, s, m, h, d. Example: '5m' for 5 minutes", 
            durationStr));
```

ksml-data-jsonschema/src/main/java/io/axual/ksml/data/notation/jsonschema/JsonSchemaNotation.java
```diff
@@ -0,0 +1,156 @@
...
+    private static final String DESERIALIZATION_ERROR_MSG = " message could not be deserialized from topic ";
+    private static final String SERIALIZATION_ERROR_MSG = " message could not be serialized to topic ";
+    private final SerdeType serdeType;
+    private final NativeDataObjectMapper nativeMapper;
+    private final Map<String, ?> serdeConfigs;
+    private final RegistryClient client;
+
+    public JsonSchemaNotation(String name, SerdeType type, NativeDataObjectMapper nativeMapper, Map<String, ?> configs, RegistryClient client) {
+        super(name, ".json", DEFAULT_TYPE, new JsonDataObjectConverter(), new JsonSchemaLoader(), nativeMapper, new JsonDataObjectMapper(false));
+        this.serdeType = type;
```
KasparMetsa, line 63: Would it make sense here to validate that schema registry URL is present for schema based serdes?
```
    String urlKey = type == SerdeType.CONFLUENT ? "schema.registry.url" : "apicurio.registry.url";
    if (!configs.containsKey(urlKey) || configs.get(urlKey) == null || 
            configs.get(urlKey).toString().trim().isEmpty()) {
        throw new ConfigException("Schema registry URL not configured. " +
            "Please set '" + urlKey + "' in the notation configuration");
    }
```

KasparMetsa (changes requested):

ksml/src/main/java/io/axual/ksml/definition/FunctionDefinition.java
jeroenvandisseldorp: Forcing the dev to use `expression` is what I just removed. This way, you are allowed to use `return` statements in your code without the `expression` statement.

ksml-data/src/main/java/io/axual/ksml/data/schema/TupleSchema.java
jeroenvandisseldorp: Added

ksml/src/main/java/io/axual/ksml/parser/DurationParser.java
jeroenvandisseldorp: Added

ksml-data-jsonschema/src/main/java/io/axual/ksml/data/notation/jsonschema/JsonSchemaNotation.java
jeroenvandisseldorp: Makes sense only if the serdes themselves would not do this validation already, and we want to give better feedback to the user than the serdes do. But these checks are done by the serdes, and their error messages are clear, so let's not duplicate the validation in KSML, as that will lead to inconsistencies and/or maintenance overhead.

ksml/src/main/java/io/axual/ksml/definition/FunctionDefinition.java
KasparMetsa: Thank you, understood

KasparMetsa (approved):

PR #284, opened 2025-08-21
Allow generator to return multiple messages

### LINKED ISSUE #283: Allow generators to return lists of messages
Problem
======

When producing messages, producers call upon `generator` functions to return newly generated messages. At the moment this can only be done one at a time. In some cases, however, it is more convenient to have generators return more than one message.

Solution
======

The solution would be to allow generators to return lists of messages, which would then represent "the next x messages to be produced". This support can easily be added by modifying the allowed `resultType` of generators to "either tuple or list of tuples".

jeroenvandisseldorp:
Fixes #283


ksml-runner/src/main/java/io/axual/ksml/runner/producer/ExecutableProducer.java
```diff
@@ -206,33 +208,58 @@ private List<Pair<byte[], byte[]>> generateBatch() {
     }
 
     private Pair<DataObject, DataObject> generateMessage() {
-        final var result = generator.apply();
-        if (result instanceof DataTuple tuple && tuple.elements().size() == 2) {
-            var key = tuple.elements().get(0);
-            var value = tuple.elements().get(1);
+        // Get a message for the queue of generated messages, or generate new messages first and then fetch
```
KasparMetsa, line 211: I think I found an edge case- let's imagine a buggy generator that accidentally does this:

```python
  def generate_messages():
      result = []
      for i in range(1000000):  # Accidentally creating a million messages
          result.append((f"key{i}", f"value{i}"))
      return result
```

Result: KSML application uses all available memory and crashes.

If we set an upper limit like this, only 10,000 messages are queued:

```java
private static final int MAX_QUEUE_SIZE = 10000;
  private boolean queueOverflowWarningIssued = false;

  private Pair<DataObject, DataObject> generateMessage() {
      if (messageQueue.isEmpty()) {
          final var newMessages = generateMessages();

          // NEW: Check queue size limit
          if (newMessages.size() > MAX_QUEUE_SIZE) {
              if (!queueOverflowWarningIssued) {
                  log.warn("Generator returned {} messages, exceeding limit of {}. Truncating.",
                           newMessages.size(), MAX_QUEUE_SIZE);
                  queueOverflowWarningIssued = true;
              }
              // Only take what we can handle
              messageQueue.addAll(newMessages.subList(0, MAX_QUEUE_SIZE));
          } else {
              messageQueue.addAll(newMessages);
          }
      }
      return messageQueue.isEmpty() ? null : messageQueue.removeFirst();
  }
```

KasparMetsa (changes requested):

KasparMetsa (approved):

Issue #338, opened 2025-09-18
Add methods to have DataSchema.equals and DataSchema.isAssignableFrom return a list of incompatibilities


jeroenvandisseldorp:
When recursing through DataSchema (and DataType?), build up a list of human readable incompatibilities, so that developers quickly understand why two schemas are not equal, or one is not assignable to the other. This speeds up schema debugging during development.

PR #375, opened 2025-10-09
338 refactor assignablefrom and equals

jeroenvandisseldorp:
* Added equals() interface for data objects and data types.
* Added Equal class to represent whether two objects are equal. If not, Equal contains the reason why they're not equal.
* Added Assignable class to represent whether an object is assignable from another. If not, Assignable contains the reason why it's not assignable.
* Util classes help in creating uniform error messages.

Implements #338

### DIFF
--- a/ksml-data-avro/src/test/java/io/axual/ksml/data/notation/avro/AvroSchemaMapperTest.java
+++ b/ksml-data-avro/src/test/java/io/axual/ksml/data/notation/avro/AvroSchemaMapperTest.java
@@ -465,4 +465,69 @@ class AvroSchemaMapperTest {
             }
         }
     }
+
+    @Test
+    @DisplayName("Optional AVRO field without explicit null default")
+    void testOptionalFieldWithoutNullDefault() {
+        // Test optional field with no default value, using JSON to create a union type without specifying a default
+        String schemaJson = """
+                {
+                  "type": "record",
+                  "name": "Test",
+                  "fields": [
+                    {
+                      "name": "optionalField",
+                      "type": ["null", "string"]
+                    }
+                  ]
+                }
+                """;
+        Schema avroSchema = new Schema.Parser().parse(schemaJson);
+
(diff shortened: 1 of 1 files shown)


tonvanbart:
I have my doubts about the usage pattern of the flags?
```java
// Example: Comparing StructType from Avro schema vs Protobuf schema
Flags flags = new Flags(
    DataTypeFlags.IGNORE_DATA_TYPE_CONTAINER_CLASS,  // Different Java representations
    DataTypeFlags.IGNORE_STRUCT_TYPE_SCHEMA          // Ignore embedded schema differences
);

DataType avroType = avroSchemaMapper.toDataType(avroSchema);
DataType protobufType = protobufSchemaMapper.toDataType(protobufSchema);

Equal result = avroType.equals(protobufType, flags);
```
It seems to me that at runtime it would not be clear which types `avroType` and `protobufType` actually would be, the names would be different and the instances created somewhere else in the code at runtime. Any caller would have to work out the flags at runtime, and forgetting a (or more) flags would lead to runtime bugs which would not be straight apparent and may be hard to track.
A better solution would have each Type/Field etc keep track of its _own_ capabilities (tag, ordering etc) and hide the flags from the caller. Maybe something along the following lines:

1. Caller --> fieldA.equals(fieldB)
2. fieldA --> fieldB.provideFlags(fieldA)  // fieldB now has both types
3. fieldB --> FlagsUtility.getFlags(typeA, typeB) 
4. fieldB --> fieldA: flags
5. fieldA --> EqualsUtility.isEqual(fieldA, fieldB, flags) 

This would have the following advantages:

- caller does not have to provide flags, flags can't be forgotten
- every type only keeps track of its own "flavor" for lack of a better word - Proto, Avro, JSON etc
- knowledge of all the compatibilities is in a single place (called `FlagsUtility` in the example above); easy to change and to track down errors

I've already used Claude to create a compatibility matrix with flags needed for the different combinations: https://github.com/tonvanbart/ksml/blob/claude-datafield-analysis/FORMAT_COMPATIBILITY_MATRIX.md

Issue #362, opened 2025-10-02
Integration Tests improvements - XML and CSV ITs, Partitioning IT


KasparMetsa:
- Refine XML and CSV ITs to use standard parsing libraries on the resulting data and compare.
  - Like using an XML Parser to create DOM object and use that to compare to expected data
- Write IT for partitioning(repartition and streamPartitioner) because TopologyTestDriver doesn't expose partition assignments, so we can't directly verify the custom partitioner logic in processing definition like this: repartition-example-processor.yaml
  - When this is done, can remove this test as it only partially validates the definition: `ksml/ksml/src/test/java/io/axual/ksml/docs/examples/reference/operations/RepartitionExampleTest.java`

PR #382, opened 2025-10-20
Improve XML and CSV tests. Partitioning IT test

KasparMetsa:
* Added Partitioning IT test. More details: https://github.com/Axual/ksml/issues/362#issuecomment-3420692517
* Use standard parsing libraries for CSV and XML validation in CSV and XML tests

  - Replaced weak string-based validation (contains, split, character counting) with robust parsing using Apache Commons   CSV and DOM parsers.
  - Changes:
    - Integration tests: Use CSV parser and DOM parser to validate structure
    - Unit tests: Use exact equality checks instead of substring matching

### DIFF
--- a/ksml-integration-tests/src/test/java/io/axual/ksml/integration/RepartitionIT.java
+++ b/ksml-integration-tests/src/test/java/io/axual/ksml/integration/RepartitionIT.java
@@ -124,7 +124,20 @@ class RepartitionIT {
         // Validate we have records
         assertThat(allRecords).as("Should have collected repartitioned records").isNotEmpty();
 
-        // Validate partitioning logic for each user
+        // Verify partitioning and transformations
+        verifyPartitioningLogic(userPartitions);
+        verifyKeyTransformationAndValueEnrichment(allRecords);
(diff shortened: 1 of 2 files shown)


ksml-integration-tests/src/test/java/io/axual/ksml/integration/RepartitionIT.java
```diff
@@ -0,0 +1,218 @@
...
+                userPartitions.computeIfAbsent(userId, k -> new ArrayList<>()).add(partition);
+
+                log.info("Record: key={}, partition={}, value preview={}",
+                         userId, partition, record.value().substring(0, Math.min(100, record.value().length())));
+            });
+        }
+
+        // Validate we have records
+        assertThat(allRecords).as("Should have collected repartitioned records").isNotEmpty();
+
```
tonvanbart: Looks OK, the method is a bit long, actually there are 2 groups of assertions here; one about partitioning, and one about key transformation and value enrichment. Maybe in future see if we can split those out in separate methods (for readability).

PR #422, opened 2025-11-24
MVP for externalized Python

tonvanbart:
This PR adds support for externalized Python and syntax highlighting and code completion for KSML Python files:

- `code:` snippets can reference external `file:`
- `globalCode:` snippets can reference external `file:`
- customer modules can be put in a configurable directory and will be available for import
- Python interface stub module for code completion and syntax highlighting in Python editors
- tests for the above
- added documentation page on external Python files


.github/workflows/build-snapshot-docs.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
```
KasparMetsa: When testing this docker compose setup:
[local-docker-compose-setup.zip](https://github.com/user-attachments/files/23757179/local-docker-compose-setup.zip)
This will crash with error:
```
org.graalvm.polyglot.PolyglotException: ModuleNotFoundError: No module named 'random'
```
The error will go away if we remove "import random". 
This is a bug because "import random" and other Python standard library imports are used in many existing code blocks before this PR.

**Summary of the Bug**

When pythonModulePath is set in ksml-runner.yaml, this PR creates a restricted file system that only allows access to that specific path. This blocks Python from accessing its standard library.

  The problematic code in PythonContext.java:
```
  FileSystem restricted = FileSystem.newCompositeFileSystem(
      denyAllAccess,  // Default: deny everything
      FileSystem.Selector.of(modulesFileSystem, path -> path.startsWith(modulePath))
  );
```
  This means:
  - /ksml/order_utils.py → Allowed
  - /opt/graal/lib/python/random.py → DENIED (not in /ksml)

Please find here Claude Code suggestions to solve this bug: 

[PR_422_PYTHONMODULEPATH_BUG.md](https://github.com/user-attachments/files/23757195/PR_422_PYTHONMODULEPATH_BUG.md)

KasparMetsa (changes requested):

KasparMetsa (commented):
Good job

ksml/src/main/java/io/axual/ksml/python/LoggerBridge.java
```diff
@@ -36,6 +36,17 @@ public PythonLogger getLogger(String loggerName) {
      * The PythonLogger has less ambiguity by removing the multiple object and exception arguments
      */
     public static class PythonLogger {
+        /** The instances wrapped Slf4J {@link org.slf4j.Logger}. */
```
KasparMetsa: Grammar error: "The instances wrapped" should be "The instance's wrapped".

Suggested fix:
```java
/** The instance's wrapped Slf4J {@link org.slf4j.Logger}. */
```

ksml/src/test/java/io/axual/ksml/testutil/KSMLTest.java
```diff
@@ -36,11 +36,14 @@
...
     String NO_SCHEMAS = "";
+    String NO_MODULES = "";
 
     /** Classpath relative reference to the pipeline definition under test. */
     String topology();
 
     /** Optional classpath relative reference to a directory with AVRO schema definitions. */
     String schemaDirectory() default NO_SCHEMAS;
 
+    String modulesDirectory() default NO_MODULES;
```
KasparMetsa: Naming inconsistency: this parameter is called `modulesDirectory` but the runtime config uses `pythonModulePath`.

This could confuse users who see `modulesDirectory` in tests but need to use `pythonModulePath` in their ksml-runner.yaml. Consider renaming to `pythonModulePath` for consistency, or document the difference clearly in the docs.

ksml-runner/src/main/java/io/axual/ksml/runner/config/KSMLConfig.java
```diff
@@ -29,6 +29,15 @@
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import java.io.IOException;
```
KasparMetsa: These imports appear to be duplicated elsewhere in the file (check lines 53-58). IDEs will flag this as a warning. Consider running "optimize imports" before submitting.

ksml/src/test/resources/pipelines/test_filter_module.py
```diff
@@ -0,0 +1,19 @@
...
+    if val == None:
+        # log.warn("No value in message with key={}", key)  <==== NEED TO FIX THIS, NOT PRESENT IN IMPORT MODULE!
+        return False
+    if val["color"] != "blue":
+        # log.warn("Unknown color: {}", val["color"])
+        return False
+    return True
+
+def is_red(val):
+    """ Filter values in the stream with attribute 'color' having value 'blue'"""
```
KasparMetsa:
```suggestion
    """ Filter values in the stream with attribute 'color' having value 'red'"""
```

ksml/src/main/java/io/axual/ksml/python/PythonContext.java
```diff
@@ -25,6 +25,10 @@
 import io.axual.ksml.metric.Metrics;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+
+import org.apache.commons.lang3.StringUtils;
+import org.graalvm.polyglot.*;
```
KasparMetsa: This import makes the individual imports below redundant.

KasparMetsa (changes requested):

docs/getting-started/editing-python-code.md
```diff
@@ -0,0 +1,243 @@
+# Editing Python Code
```
KasparMetsa: Used Claude to find a problem with syntax like this:
```
functions:
  # Import inside code: (runs on every message - inefficient but works)
  log_action_count:
    type: forEach
    code: |
      from processor_functions import log_user_action_count
      log_user_action_count(key, value, log)
```

## The Problem

When using `code:` for imports, the import statement executes **on every message**:

```python
# Generated code when using code: for import
def log_action_count(key, value):
    from processor_functions import log_user_action_count  # RUNS EVERY CALL
    log_user_action_count(key, value, log)
```

When using `globalCode:`, the import executes **once at startup**:

```python
# Generated code when using globalCode: for import
from processor_functions import log_user_action_count  # RUNS ONCE

def log_action_count(key, value):
    log_user_action_count(key, value, log)
```

Here is full description, proof and possible solutions by Claude

[PR_422_IMPORT_OVERHEAD_ANALYSIS.md](https://github.com/user-attachments/files/23954124/PR_422_IMPORT_OVERHEAD_ANALYSIS.md)

Here is benchmark by Clade to prove the overhead:

[benchmark_import.py](https://github.com/user-attachments/files/23954132/benchmark_import.py)

When I ran it, I got this output:

[benchmark.txt](https://github.com/user-attachments/files/23954162/benchmark.txt)

KasparMetsa (changes requested):

KasparMetsa (changes requested):

docs/getting-started/editing-python-code.md
```diff
@@ -0,0 +1,243 @@
...
+
+??? info "Complete ksml.py module (click to expand)"
+
+    ```python
+    {% include "../../ksml/src/test/resources/ksml.py" %}
+    ```
+
+## Complete Example: Fraud Detection
+
+For a comprehensive example showing all these patterns together, see the **Fraud Detection use case** in `docs-examples/use-cases/fraud-detection/`:
```
KasparMetsa: Could we add a link to the fraud detection use case please here?

KasparMetsa (changes requested):

ksml/src/test/resources/pipelines/test-filter-module-import.yaml
```diff
@@ -0,0 +1,33 @@
...
+    topic: ksml_sensordata_avro
+    keyType: string
+    valueType: avro:SensorData
+  sensor_filtered:
+    topic: ksml_sensordata_filtered
+    keyType: string
+    valueType: avro:SensorData
+
+functions:
+  module_import:
```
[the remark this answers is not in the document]
tonvanbart: ? This is as Jeroen suggested it, a dummy intialization function

jeroenvandisseldorp (approved):

KasparMetsa (approved):

PR #450, opened 2026-02-16
Fix avro state stores

tonvanbart:
Added tests for AVRO state stores; updated test tooling; fixed regression in AVRO state stores

### DIFF
--- a/ksml/src/test/java/io/axual/ksml/operation/KSMLStateStoreTest.java
+++ b/ksml/src/test/java/io/axual/ksml/operation/KSMLStateStoreTest.java
@@ -30,10 +30,12 @@ import org.apache.avro.generic.GenericRecord;
 import org.apache.kafka.streams.TestInputTopic;
 import org.apache.kafka.streams.TopologyTestDriver;
 import org.apache.kafka.streams.state.KeyValueStore;
+import org.apache.kafka.streams.state.ValueAndTimestamp;
 import org.apache.kafka.streams.state.WindowStore;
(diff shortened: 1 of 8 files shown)


ksml/src/test/java/io/axual/ksml/operation/KSMLStateStoreTest.java
```diff
@@ -110,6 +114,92 @@ void testJoinTimestamped() {
...
+                .value("80")
+                .build().toRecord());
+        sensorIn.pipeInput("sensor2", SensorData.builder()
+                .city("Utrecht")
+                .type(SensorData.SensorType.TEMPERATURE)
+                .unit("C")
+                .value("26")
+                .build().toRecord());
+
+        KeyValueStore<Object, Object> lastSensorDataStore = topologyTestDriver.getKeyValueStore("last_sensor_data_store");
```
KasparMetsa: `getKeyValueStore()` silently unwraps the timestamp and returns just the value. This means the test passes but never actually verifies that the timestamped store works as a timestamped store. Using `getTimestampedKeyValueStore()` ensures the test exercises the real API, catching regressions in timestamp handling. The new Avro variant already does this - the old test should match.

```java
var lastSensorDataStore = topologyTestDriver.getTimestampedKeyValueStore("last_sensor_data_store");
```

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #505, opened 2026-04-09
Ksml pipeline testing

tonvanbart:
Added module ksml-test-runner which gives an extra entry point in the KSML Docker image. The entry point reads one or more test definitions in YAML format, runs the tests (using Kafka's `TopologyTestDriver`) and reports on the results. 
This makes it possible to test KSML pipelines without having to use Java.
The PR also adds documentation on the feature: see https://tonvanbart.github.io/ksml/DEVELOPMENT/getting-started/testing-your-pipeline/
When building KSML, a JSON schema is generated for the test YAML, similar to what happens for the runner definition and the pipeline definition.


ksml-test-runner/src/main/java/io/axual/ksml/testrunner/KSMLTestRunner.java
```diff
@@ -0,0 +1,322 @@
...
+     * Files are accepted as-is; directories are walked recursively for *.yaml / *.yml.
+     */
+    static List<Path> collectTestFiles(List<File> inputs) throws IOException {
+        var files = new ArrayList<Path>();
+        for (var input : inputs) {
+            var path = input.toPath();
+            if (Files.isDirectory(path)) {
+                try (var stream = Files.walk(path)) {
+                    stream.filter(Files::isRegularFile)
+                            .filter(KSMLTestRunner::isYamlFile)
```
KasparMetsa, line 264: Could we add a filter here to check if the file has "test:" root. Something like:

```java
.filter(KSMLTestRunner::hasTestRoot)
```
```
    static boolean hasTestRoot(Path path) {
        try {
            var root = YAMLObjectMapper.INSTANCE.readValue(Files.readString(path), JsonNode.class);
            var present = root != null && root.has("test");
            if (!present) log.debug("Skipping {}: no 'test' root element", path);
            return present;
        } catch (IOException e) {
            log.debug("Skipping {}: could not read YAML ({})", path, e.getMessage());
            return false;
        }
    }
```

Currently when the directory contains files like this:

Image

And I run a command like this:
```shell
java -Djava.security.manager=allow \
  -jar /Users/km/dev/ksml/ksml-test-runner/target/ksml-test-runner-1.3.0-SNAPSHOT.jar \
  /Users/km/dev/ksml/ksml-pr505-demo/
```

The test output is:

```
=== KSML Test Results ===

  PASS  Enrich adds amount_with_tax field at 21%
  PASS  Branch routes orders by category to separate topics
  FAIL  Intentionally failing assertion
        AssertionError: Expected 99, got 1
  PASS  Double-amount pipeline doubles every order's amount
  PASS  Filter keeps only orders with amount > 50
  ERROR add-tax-field.yaml
        Invalid test definition: Missing required 'test' root element in /Users/km/dev/ksml/ksml-pr505-demo/pipelines/add-tax-field.yaml
  ERROR branch-by-category.yaml
        Invalid test definition: Missing required 'test' root element in /Users/km/dev/ksml/ksml-pr505-demo/pipelines/branch-by-category.yaml
  ERROR double-value.yaml
        Invalid test definition: Missing required 'test' root element in /Users/km/dev/ksml/ksml-pr505-demo/pipelines/double-value.yaml
  ERROR filter-high-value.yaml
        Invalid test definition: Missing required 'test' root element in /Users/km/dev/ksml/ksml-pr505-demo/pipelines/filter-high-value.yaml

4 passed, 1 failed, 4 errors
```

After this fix, the output is:

```
=== KSML Test Results ===

  PASS  Enrich adds amount_with_tax field at 21%
  PASS  Branch routes orders by category to separate topics
  FAIL  Intentionally failing assertion
        AssertionError: Expected 99, got 1
  PASS  Double-amount pipeline doubles every order's amount
  PASS  Filter keeps only orders with amount > 50

4 passed, 1 failed, 0 errors
```

KasparMetsa (changes requested):

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/KSMLTestRunner.java
tonvanbart: Let's discuss, it's currently intentional to signal ERROR (instead of FAIL) in the case of a malformed test file. The proposed solution would silently ignore them which might be surprising for test authors, especially if there are a lot of tests. WDYT?

tonvanbart:
Changed testing YAML dialect as discussed, reopen PR

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/KSMLTestRunner.java
KasparMetsa: Ok sure

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/AssertionRunner.java
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * Run all assertion blocks and return the result.
+     *
+     * @param assertBlocks the assertion blocks to execute
+     * @param suiteName    the suite-level name for reporting
+     * @param testName     the test display label for reporting
+     * @return the test result
+     */
+    public TestResult runAssertions(List<AssertBlock> assertBlocks, String suiteName, String testName) {
```
KasparMetsa, line 67: When a test has two `assert` blocks that both reference the same output stream via `on:`, the second block always receives `records = []` and any assertion inside it fails.

The root cause is in `AssertionRunner.collectOutputRecords`: it calls `readRecordsToList()` on the `TopologyTestDriver` queue every time it is invoked. That call permanently drains the queue. The first block gets the records; the second block calls the same method, finds the queue empty, and injects an empty list into Python.

Steps to reproduce:
1. Create `ksml/ksml-test-runner/src/test/resources/two-assert-blocks.yaml`

```yaml
name: "Two assert blocks on the same stream"
definition: pipelines/test-filter.yaml
schemaDirectory: schemas

streams:
  sensor_source:
    topic: ksml_sensordata_avro
    keyType: string
    valueType: "avro:SensorData"
  sensor_filtered:
    topic: ksml_sensordata_filtered
    keyType: string
    valueType: "avro:SensorData"

tests:
  two_blocks_same_stream:
    description: "Two assert blocks both reference sensor_filtered"
    produce:
      - to: sensor_source
        messages:
          - key: "sensor-1"
            value:
              name: "sensor-1"
              timestamp: 1000
              value: "25.0"
              type: "TEMPERATURE"
              unit: "celsius"
              color: "blue"
          - key: "sensor-2"
            value:
              name: "sensor-2"
              timestamp: 2000
              value: "60.0"
              type: "HUMIDITY"
              unit: "percent"
              color: "red"

    assert:
      # Block 1 — drains the topic queue
      - on: sensor_filtered
        code: |
          assert len(records) == 1, f"Expected 1 filtered record, got {len(records)}"

      # Block 2 — topic queue is now empty; records is []
      - on: sensor_filtered
        code: |
          assert len(records) == 1, f"Expected 1 record in second block, got {len(records)}"
          assert records[0]["key"] == "sensor-1", f"Expected key 'sensor-1', got {records[0]['key']}"
```
Run something like:
```shell
  java -Djava.security.manager=allow \
    -jar ksml-test-runner/target/ksml-test-runner-1.3.0-SNAPSHOT.jar \
    ksml-test-runner/src/test/resources/two-assert-blocks.yaml
```

## Expected output

Both blocks see the same record, both pass:

```
=== KSML Test Results ===

  PASS  Two assert blocks on the same stream › Two assert blocks both reference sensor_filtered

1 passed, 0 failed, 0 errors
```

## Actual output (what actually happens)

```
=== KSML Test Results ===

  FAIL  Two assert blocks on the same stream › Two assert blocks both reference sensor_filtered
        AssertionError: Expected 1 record in second block, got 0

0 passed, 1 failed, 0 errors
```

Block 1 passes. Block 2 fails because `records` is empty.

Suggested fix(screenshot+file)

Image

[AssertionRunner.java](https://github.com/user-attachments/files/27627635/AssertionRunner.java)

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/AssertionRunner.java
```diff
@@ -0,0 +1,236 @@
...
+        var pythonContext = new PythonContext(PythonContextConfig.builder().build());
```
KasparMetsa: Code review from Claude:
PythonContext is never closed (GraalVM resource leak).
`PythonContext` wraps a GraalVM `org.graalvm.polyglot.Context`, which implements `AutoCloseable`. Failing to close it leaks native memory and threads. In a test suite with many tests, this accumulates.

More details:
[fix-issue-python-context-leak.md](https://github.com/user-attachments/files/27628541/fix-issue-python-context-leak.md)

Fixed files attached:

[AssertionRunner.java](https://github.com/user-attachments/files/27628578/AssertionRunner.java)

[PythonContext.java](https://github.com/user-attachments/files/27628583/PythonContext.java)

[TestDataProducer.java](https://github.com/user-attachments/files/27628593/TestDataProducer.java)

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/TestExecutionContext.java
```diff
@@ -0,0 +1,149 @@
...
+        } catch (IOException | io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException e) {
+            throw new TestDefinitionException(
+                    "Failed to register schema for subject '" + subject + "': " + e.getMessage(), e);
+        }
+    }
+
+    /**
+     * Reset the execution context for test isolation.
+     */
+    public void cleanup() {
```
KasparMetsa, line 144: This is probably a non-issue in 99.9% of the cases, so please feel free to ignore it, but what Claude found here during review could still be interesting to read: 

[issue-notation-library-not-cleaned-up.md](https://github.com/user-attachments/files/27628929/issue-notation-library-not-cleaned-up.md)

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/KSMLTestRunner.java
```diff
@@ -0,0 +1,342 @@
...
+        }
+        var url = getClass().getClassLoader().getResource(directoryName);
+        if (url != null && "file".equals(url.getProtocol())) {
+            return url.getPath();
+        }
+        var absolute = Path.of(directoryName);
+        if (Files.isDirectory(absolute)) {
+            return absolute.toAbsolutePath().toString();
+        }
+        return directoryName;
```
KasparMetsa: Steps to reproduce a small bug:
1. Create `ksml/ksml-test-runner/src/test/resources/bad-schema-dir-test.yaml`
```yaml
name: "Bad schema directory test"
definition: pipelines/test-filter.yaml
schemaDirectory: does-not-exist

streams:
  sensor_source:
    topic: ksml_sensordata_avro
    keyType: string
    valueType: "avro:SensorData"
  sensor_filtered:
    topic: ksml_sensordata_filtered
    keyType: string
    valueType: "avro:SensorData"

tests:
  filter_test:
    description: "Should fail with a clear error about the missing schema directory"
    produce:
      - to: sensor_source
        messages:
          - key: "sensor-1"
            value:
              name: "sensor-1"
              timestamp: 1000
              value: "25.0"
              type: "TEMPERATURE"
              unit: "celsius"
              color: "blue"
    assert:
      - on: sensor_filtered
        code: |
          assert len(records) == 1
```

2. Run:
```bash
java -Djava.security.manager=allow \
  -jar ksml-test-runner/target/ksml-test-runner-1.3.0-SNAPSHOT.jar \
  ksml-test-runner/src/test/resources/bad-schema-dir-test.yaml
```
3. Actual error:
```
ERROR bad-schema-dir-test › Should fail with a clear error about the missing schema directory
      Invalid test definition: Schema file not found for registry entry: does-not-exist/SensorData.avsc
                               (topic=ksml_sensordata_avro, type=avro:SensorData)
```
4. Expected error:
```
ERROR bad-schema-dir-test › <parse>
      Invalid test definition: Directory 'does-not-exist' not found
                               (resolved from ksml-test-runner/src/test/resources/bad-schema-dir-test.yaml)
```

Suggested Fixes by Claude with reasons why:
**Fix#1**
`resolveDirectory` tries three locations in order:

1. Next to the test file — `<test-file-folder>/schemas`
2. On the classpath (for schemas bundled inside the JAR)
3. As an absolute path — `/schemas`

If all three fail, the folder does not exist.

Currently, when all three searches failed the method returns the raw string the user typed - `"schemas"` - as if it had found something. Then the string is passed into the runner, which tries to open schema files inside it. This error points at the schema file, not at the real problem - the `schemas` folder itself never existed.

```java
// Before — silently falls through
return directoryName;

// After — throws with a clear message
throw new TestDefinitionException(
    "Directory '" + directoryName + "' not found (resolved from " + testFile + ")");
```
**Fix#2**
`resolveDirectory` was called outside the try-catch block, so the new exception crashed the whole runner as an unhandled exception instead of producing a clean test result. Moving those two calls inside the try block made the exception land in the existing `TestDefinitionException` handler, which formats it as a proper error result.

```java
// Before — outside the try-catch, unhandled if it throws
} catch (...) { ... }

var schemaDirectory = resolveDirectory(testFile, suite.schemaDirectory());   // crash here
var modulesDirectory = resolveDirectory(testFile, suite.moduleDirectory());

// After — inside the try-catch, formatted as a test ERROR
try {
    suite = parser.parse(testFile);
    definitionYaml = readDefinitionContent(testFile, suite.definition());
    schemaDirectory = resolveDirectory(testFile, suite.schemaDirectory());   // caught here
    modulesDirectory = resolveDirectory(testFile, suite.moduleDirectory());
} catch (TestDefinitionException e) {
    return List.of(TestResult.error(..., "Invalid test definition: " + e.getMessage()));
}
```

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/AssertBlock.java
```diff
@@ -0,0 +1,58 @@
...
+                examples = {"my_store"})
+        List<String> stores,
+
+        @JsonSchema(description = "Python assertion code to execute", required = true)
+        String code
+) {
+    /**
+     * Validate that the assert block has at least an {@code on} reference or {@code stores}.
+     */
+    public void validate() {
```
KasparMetsa, line 53: Steps to reproduce a small edge case bug:
1. Create `ksml-test-runner/src/test/resources/empty-stores-test.yaml`

```yaml
name: "Empty stores list test"
definition: pipelines/test-filter.yaml
schemaDirectory: schemas

streams:
  sensor_source:
    topic: ksml_sensordata_avro
    keyType: string
    valueType: "avro:SensorData"
  sensor_filtered:
    topic: ksml_sensordata_filtered
    keyType: string
    valueType: "avro:SensorData"

tests:
  empty_stores_passes_validation:
    description: "stores is an empty list — should be caught at parse time but is not"
    produce:
      - to: sensor_source
        messages:
          - key: "sensor-1"
            value:
              name: "sensor-1"
              timestamp: 1000
              value: "25.0"
              type: "TEMPERATURE"
              unit: "celsius"
              color: "blue"
    assert:
      - stores: []
        code: |
          assert len(records) == 1, f"Expected 1 record, got {len(records)}"
```

Note please that `stores: []` is an empty list by user mistake.

2. Run the test

```bash
java -Djava.security.manager=allow \
  -jar ksml-test-runner/target/ksml-test-runner-1.3.0-SNAPSHOT.jar \
  ksml-test-runner/src/test/resources/empty-stores-test.yaml
```

3. Error:
## Expected output

The empty `stores` list should be caught when the YAML is parsed, before any test runs:

```
ERROR empty-stores-test › <parse>
      Invalid test definition: Assert block must have at least 'on' or a non-empty 'stores' list
```

## Actual output

Validation passes. The Python code runs with no variables injected. The first reference to `records` crashes:

```
ERROR Empty stores list test › stores is an empty list — should be caught at parse time but is not
      NameError: name 'records' is not defined
```

The error points at the Python code. There is no mention of `stores: []` or the YAML file. User would read this and start looking for a problem in their assertion code, when the real mistake is a typo in the YAML.

## The fix

Also reject an empty list:

```java
public void validate() {
    var hasOn = on != null;
    var hasStores = stores != null && !stores.isEmpty();
    if (!hasOn && !hasStores) {
        throw new TestDefinitionException(
            "Assert block must have at least 'on' or a non-empty 'stores' list");
    }
}
```

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/TestDataProducer.java
```diff
@@ -0,0 +1,229 @@
...
+            } else {
+                inputTopic.pipeInput(msg.key(), msg.value());
+            }
+        }
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private void produceGeneratedMessages(ProduceBlock block, StreamDefinition stream, long count) {
+        var generatorMap = block.generator();
+        var generatorName = (String) generatorMap.getOrDefault("name", "test_generator");
```
KasparMetsa: Steps to reproduce small edge case bug:

1. Create the test file `ksml-test-runner/src/test/resources/generator-bad-field-type-test.yaml`
```yaml
name: "Generator with wrong field type"
definition: pipelines/test-passthrough.yaml

streams:
  input:
    topic: input_topic
    keyType: string
    valueType: string
  output:
    topic: output_topic
    keyType: string
    valueType: string

tests:
  wrong_name_type:
    description: "name field is a number instead of a string"
    produce:
      - to: input
        generator:
          name: 42              # <-- number instead of a string
          code: |
            result = [("key-1", "value-1")]
          expression: result
        count: 1

    assert:
      - on: output
        code: |
          assert len(records) == 1
```
2. Run the test

```bash
java -Djava.security.manager=allow \
  -jar ksml-test-runner/target/ksml-test-runner-1.3.0-SNAPSHOT.jar \
  ksml-test-runner/src/test/resources/generator-bad-field-type-test.yaml
```

3. Error:
## Expected output

A clear error naming the field and the wrong type:

```
ERROR Generator with wrong field type › name field is a number instead of a string
      Invalid test definition: Generator field 'name' must be a string, got: Integer
```

---

## Actual output

```
ERROR Generator with wrong field type › name field is a number instead of a string
      class java.lang.Integer cannot be cast to class java.lang.String
      (java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap')
```

This is a raw Java error. It says nothing about which field is wrong, what value was found, or which test file caused it.

## The fix

Replace the direct casts with a helper that checks the type first:

```java
private static String getStringOrDefault(Map<String, Object> map, String key, String defaultValue) {
    var val = map.get(key);
    if (val == null) return defaultValue;
    if (!(val instanceof String)) {
        throw new TestDefinitionException(
            "Generator field '" + key + "' must be a string, got: " + val.getClass().getSimpleName());
    }
    return (String) val;
}
```

Use it for all four fields:

```java
var generatorName = getStringOrDefault(generatorMap, "name", "test_generator");
var globalCode    = getStringOrDefault(generatorMap, "globalCode", "");
var code          = getStringOrDefault(generatorMap, "code", "");
var expression    = getStringOrDefault(generatorMap, "expression", "result");
```

ksml-test-runner/README.md
```diff
@@ -0,0 +1,292 @@
...
+- Schema-less notations (`string`, `long`, `int`, `boolean`, `bytes`, `json`, `binary`, `xml`, etc.)
+  may appear unqualified.
+- Schema-bearing notations (`avro`, `confluent_avro`, `apicurio_avro`, `protobuf`, `json_schema`,
+  `csv`) **must** be qualified with a schema name (e.g., `avro:SensorData`). The named schema is
+  loaded from `schemaDirectory` and registered in the in-memory mock registry under
+  `<topic>-key` / `<topic>-value`.
+- Bare schema-bearing notations (e.g., `confluent_avro` without `:Schema`) are rejected — the
+  test runner has no real registry to resolve them against.
+
+### Tests
```
KasparMetsa, line 99: Could we add something like:
```
!!! warning "Supported notations"
    The test runner supports: `string`, `int`, `long`, `boolean`, `bytes`, `json`, `binary`, `avro`, `confluent_avro`, `apicurio_avro`.

    `apicurio_avro` is backed by a Confluent-compatible mock registry. It works for logic-level tests, but uses Confluent wire format rather than Apicurio's.

    The following notations are **not yet supported** and will produce a runtime error if used: `xml`, `soap`, `csv`, `protobuf`, `json_schema`.
```

docs/getting-started/testing-your-pipeline.md
```diff
@@ -0,0 +1,287 @@
...
+| `keyType` | no | `string` | Key serialization type (e.g. `string`, `avro:MyKey`). |
+| `valueType` | no | `string` | Value serialization type (e.g. `string`, `avro:SensorData`). |
+
+Type strings follow KSML's standard type grammar — the same one `StreamDefinitionParser` uses in pipeline yamls:
+
+- **Schema-less notations** (`string`, `long`, `int`, `boolean`, `bytes`, `json`, `binary`, `xml`, etc.) may appear unqualified.
+- **Schema-bearing notations** (`avro`, `confluent_avro`, `apicurio_avro`, `protobuf`, `json_schema`, `csv`) **must** be qualified with a schema name (e.g., `avro:SensorData`). The named schema is loaded from `schemaDirectory` and registered in an in-memory mock registry under `<topic>-key` / `<topic>-value`. 
+- **Bare schema-bearing notations** (e.g., `confluent_avro` without `:Schema`) are rejected — the test runner has no real registry to resolve them against.
+
+### Tests
```
KasparMetsa, line 89: Could we add something like:
```
!!! warning "Supported notations"
    The test runner supports: `string`, `int`, `long`, `boolean`, `bytes`, `json`, `binary`, `avro`, `confluent_avro`, `apicurio_avro`.

    `apicurio_avro` is backed by a Confluent-compatible mock registry. It works for logic-level tests, but uses Confluent wire format rather than Apicurio's.

    The following notations are **not yet supported** and will produce a runtime error if used: `xml`, `soap`, `csv`, `protobuf`, `json_schema`.
```

docs/getting-started/testing-your-pipeline.md
```diff
@@ -0,0 +1,287 @@
...
+
+| Field | Required | Default | Description |
+|---|---|---|---|
+| `topic` | yes | | Kafka topic name. Each topic may appear in at most one stream entry. |
+| `keyType` | no | `string` | Key serialization type (e.g. `string`, `avro:MyKey`). |
+| `valueType` | no | `string` | Value serialization type (e.g. `string`, `avro:SensorData`). |
+
+Type strings follow KSML's standard type grammar — the same one `StreamDefinitionParser` uses in pipeline yamls:
+
+- **Schema-less notations** (`string`, `long`, `int`, `boolean`, `bytes`, `json`, `binary`, `xml`, etc.) may appear unqualified.
```
KasparMetsa:
```suggestion
- **Schema-less notations** (`string`, `long`, `int`, `boolean`, `bytes`, `json`, `binary`) may appear unqualified.
```

docs/getting-started/testing-your-pipeline.md
```diff
@@ -0,0 +1,287 @@
...
+- **Schema-bearing notations** (`avro`, `confluent_avro`, `apicurio_avro`, `protobuf`, `json_schema`, `csv`) **must** be qualified with a schema name (e.g., `avro:SensorData`). The named schema is loaded from `schemaDirectory` and registered in an in-memory mock registry under `<topic>-key` / `<topic>-value`.
```
KasparMetsa:
```suggestion
- **Schema-bearing notations** (`avro`, `confluent_avro`, `apicurio_avro`) **must** be qualified with a schema name (e.g., `avro:SensorData`). The named schema is loaded from `schemaDirectory` and registered in an in-memory mock registry under `<topic>-key` / `<topic>-value`. 
```

ksml-test-runner/README.md
```diff
@@ -0,0 +1,292 @@
...
+| `topic` | yes | Kafka topic name. Each topic may appear in at most one stream entry. |
+| `keyType` | no | Key type string (default `string`). |
+| `valueType` | no | Value type string (default `string`). |
+
+Type strings follow KSML's standard type grammar (the same one `StreamDefinitionParser` uses
+in pipeline yamls):
+
+- Schema-less notations (`string`, `long`, `int`, `boolean`, `bytes`, `json`, `binary`, `xml`, etc.)
+  may appear unqualified.
+- Schema-bearing notations (`avro`, `confluent_avro`, `apicurio_avro`, `protobuf`, `json_schema`,
```
KasparMetsa: For testing we don't support: protobuf, json_schema, csv

ksml-test-runner/README.md
```diff
@@ -0,0 +1,292 @@
...
+| Field | Required | Description |
+|---|---|---|
+| `topic` | yes | Kafka topic name. Each topic may appear in at most one stream entry. |
+| `keyType` | no | Key type string (default `string`). |
+| `valueType` | no | Value type string (default `string`). |
+
+Type strings follow KSML's standard type grammar (the same one `StreamDefinitionParser` uses
+in pipeline yamls):
+
+- Schema-less notations (`string`, `long`, `int`, `boolean`, `bytes`, `json`, `binary`, `xml`, etc.)
```
KasparMetsa:
```suggestion
- Schema-less notations (`string`, `long`, `int`, `boolean`, `bytes`, `json`, `binary`)
```

ksml-test-runner/README.md
```diff
@@ -0,0 +1,292 @@
...
+## Running from the command line
+
+Build the project first:
+
+```bash
+mvn clean package
+```
+
+Then run one or more test files, directories, or a mix of both. When a directory is given,
+the runner walks it recursively and picks up every `*.yaml` / `*.yml` file inside:
```
KasparMetsa:
```suggestion
the runner collects `*.yaml` / `*.yml` files from its **direct children only** (not subdirectories). To test files in nested folders, pass each subdirectory explicitly:
```

KasparMetsa (changes requested):

ksml-test-runner/README.md
```diff
@@ -0,0 +1,292 @@
...
+| `code` | yes | Python assertion code. Use `assert` statements. |
+
+\* At least one of `on` or `stores` must be specified.
+
+When `on:` is set, `records` is a list of dicts with `key`, `value`, and `timestamp` fields,
+deserialized using the referenced stream's serdes. When `stores:` is set, each store is
+available as a Python variable with the same API as in pipeline code (e.g. `store.get(key)`,
+`store.put(key, value)`).
+
+Inline `topic` fields on an assert block are not permitted — declare the stream under
```
KasparMetsa, line 144: Could we add something like:

> **Note.** When the output topic's value type is `avro:…` or `confluent_avro:…`, you must declare the output stream in `streams:` with the correct `valueType` **and** provide a `schemaDirectory` containing the matching `.avsc` file. If the stream is declared as `valueType: string` (or the field is omitted), `records[i]["value"]` will be a raw byte string (Confluent wire format: magic byte + schema ID + Avro body), and subfield indexing will fail with `TypeError: string indices must be integers`.

docs/getting-started/testing-your-pipeline.md
```diff
@@ -0,0 +1,287 @@
...
+
+| Field | Required | Description |
+|---|---|---|
+| `on` | no* | Stream key (must exist in `streams:`). The runner reads all records from the underlying topic and injects them as a `records` list variable, deserialized using the stream's serdes. |
+| `stores` | no* | List of state store names (the names declared in the pipeline's `stores:` block). Each store is injected as a Python variable. |
+| `code` | yes | Python assertion code using `assert` statements. |
+
+When `on:` is set, `records` is a list of dicts with `key`, `value`, and `timestamp` fields. When `stores:` is set, each store is available as a Python variable with the same API as in pipeline functions (e.g. `store.get(key)`, `store.put(key, value)`).
+
+### Reporting
```
KasparMetsa, line 131: Could we consider adding something like:
```
!!! note "Avro output streams must declare the correct `valueType`"
    `records[i]["value"]` is a dict only when the stream's `valueType` in `streams:` matches what the pipeline actually writes. If the pipeline writes Avro but the stream is declared as `valueType: string` (or the field is omitted), the runner falls back to `StringDeserializer`. The raw Confluent wire bytes — magic byte + 4-byte schema ID + Avro payload — are returned as a string, and indexing into `records[i]["value"]` with a field name fails with `TypeError: string indices must be integers`.

    Always declare `valueType: "avro:SchemaName"` (or `confluent_avro:SchemaName`) in `streams:` for any output topic that your pipeline serializes as Avro, and provide the matching `.avsc` file via `schemaDirectory`.
```

KasparMetsa (changes requested):

ksml-test-runner/README.md
jeroenvandisseldorp: I would prefer if we do not mention confluent_avro or apicurio_avro at all in examples, since these are only meant as "bootstrap notations", but we want people to define `avro` in the ksml-runner.yaml and not use the unparameterised versions `confluent_avro` and `apicurio_avro` in their KSML definitions.

jeroenvandisseldorp: KSML definitions are supposed to be vendor neutral, so adding these notations in definitions goes against vendor neutrality for definitions.

docs/getting-started/testing-your-pipeline.md
jeroenvandisseldorp: A notation is a way of (de)serializing data. The types you mention here are data types, not notations. So I would name them "Schema-less data types".

openspec/changes/archive/2026-05-11-parser-cleanup-and-schema/design.md
```diff
@@ -0,0 +1,265 @@
...
+        definition.validate();
+        return definition;
+    }
+    
+    private List<ProduceBlock> parseProduceBlocks(JsonNode produceArray, Path testFile) {
+        return StreamSupport.stream(produceArray.spliterator(), false)
+            .map(blockNode -> {
+                var blockFields = new FieldExtractor(blockNode, testFile);
+                var block = new ProduceBlock(
+                    blockFields.requireString("topic"),
```
jeroenvandisseldorp, line 134: Can you move parsed literals like this into a KSMLTestDSL file? See KSMLDSL for example on literals for KSML definitions.

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/TestExecutionContext.java
tonvanbart: It's interesting but I wonder if the analysis is completely correct; `populateMockRegistry()` is only called after all notations are registered, "some notations were registered, some were not" can't happen IMO. (Situation1)
Situation 2 "every run adds five more entries to the notation library": the implementation is a `put` in a Map, so I don't see how it would add.
Maybe make a separate task for this for further analysis? I don't think these are real problems, WDYT?

KasparMetsa: I agree that these are not real problems. It will be obvious once the tests don't finish or take too long.

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/TestDataProducer.java
```diff
@@ -0,0 +1,242 @@
...
+                var generated = pythonFunction.call();
+                var messages = extractKeyValuePairs(generated);
+                for (var pair : messages) {
+                    inputTopic.pipeInput(pair[0], pair[1]);
+                    totalMessages++;
+                }
+            }
+            log.debug("Generator produced {} messages to topic '{}'", totalMessages, stream.topic());
+        } catch (Exception e) {
+            log.warn("PythonContext close error producing generated messages", e);
```
KasparMetsa: Found small bug: 
- When a user's Python generator code has any runtime error (typo, undefined variable, division by zero, etc.), the user sees a test failure that says a generic unrelated message, not the actual cause in the generator.

Steps to reproduce:
1. Build the test runner:`mvn package -pl ksml-test-runner -am -DskipTests=true`
2. We use this existing pipeline file: `ksml-test-runner/src/test/resources/pipelines/test-passthrough.yaml`
3. We write this test file that will fail because it has division by zero: `ksml-test-runner/src/test/resources/generator-with-python-error.yaml`

```yaml
name: "Generator with a Python error"
definition: pipelines/test-passthrough.yaml

streams:
  input:
    topic: input_topic
    keyType: string
    valueType: string
  output:
    topic: output_topic
    keyType: string
    valueType: string

tests:
  undefined_variable:
    description: "Generator references an undefined variable"
    produce:
      - to: input
        generator:
          name: bad_gen
          code: |
            result = [("k", str(1 / 0))]
          expression: result
        count: 1
    assert:
      - on: output
        code: |
          assert len(records) == 1
```
4. We run the test file:
```shell
java -Djava.security.manager=allow \
  -jar ksml-test-runner/target/ksml-test-runner-1.3.0-SNAPSHOT.jar \
  ksml-test-runner/src/test/resources/generator-with-python-error.yaml
```
5. Actual error:
```
=== KSML Test Results ===

  FAIL  Generator with a Python error › Generator references an undefined variable
        AssertionError

0 passed, 1 failed, 0 errors
```


6.  Expected error:
```
=== KSML Test Results ===

  ERROR Generator with a Python error › Generator references an undefined variable
        KSML topology error: Error while executing function test.bad_gen : ZeroDivisionError: division by zero

0 passed, 0 failed, 1 errors
```


**Fix**

Image

Image

Files:
PythonContext.java

[PythonContext.java](https://github.com/user-attachments/files/28008684/PythonContext.java)

TestDataProducer.java

[TestDataProducer.java](https://github.com/user-attachments/files/28008698/TestDataProducer.java)

**Reason for fix**
- We let the exception propagate so runSingleTest reports it as a proper test ERROR result

KasparMetsa (changes requested):

KasparMetsa (approved):

jeroenvandisseldorp (changes requested):

PR #539, opened 2026-05-01
Add configurable `runtimeClassName` support for pods in Helm chart

### LINKED ISSUE #538: Add runtimeClassName option in KSML helm chart deployments
Currently the helm charts don't have a configurable `runtimeClassName` for the statefulset and job. This option can be used to dynamically select the container runtime for the pod. In supported K8s clusters, the pod can be deployed with stricter isolation.

xsreality:
Closes #538


packaging/helm-charts/ksml/templates/statefulset.yaml
```diff
@@ -30,6 +30,9 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "ksml.serviceAccountName" . }}
+      {{- with .Values.runtimeClassName }}
```
KasparMetsa: Currently `runtimeClassName` is between `serviceAccountName` and `securityContext`. Could we consider moving it immediately before the `nodeSelector` because `runtimeClassName` is the same kind of pod-level scheduling optional and belongs in this block.

pod-level scheduling / runtime optionals (`nodeSelector`, `affinity`, `tolerations`) are grouped together near the bottom of the pod spec

packaging/helm-charts/ksml/templates/job.yaml
```diff
@@ -35,6 +35,9 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "ksml.serviceAccountName" . }}
+      {{- with .Values.runtimeClassName }}
+      runtimeClassName: {{ . }}
```
KasparMetsa: Helm renders `{{ . }}` as a raw string into the YAML. Kubernetes accepts anything that is a DNS-1123 subdomain for a RuntimeClass name (`gvisor`, `kata-cc-tdx`, `kata-fc`, `runc-1`, …). Most of those are fine unquoted, but a value that *starts with a digit* (`{{ . }}` becomes `123runtime`, parsed as a number → YAML coercion error) or contains a colon will break rendering. Every other consumer of `with .Values.<scalar>` in this chart that ends up in a pod spec uses `| quote` for exactly this reason.


**Suggested change** (both files)

```yaml
{{- with .Values.runtimeClassName }}
runtimeClassName: {{ . | quote }}
{{- end }}
```

packaging/helm-charts/ksml/README.md
```diff
@@ -171,6 +171,7 @@ The following table lists the configurable parameters of the `ksml` chart and th
 | readinessProbe.successThreshold                    | int | `1` | Minimum consecutive successes for the probe to be considered successful after having failed. |
 | readinessProbe.timeoutSeconds                      | int | `1` | Number of seconds after which the probe times out. |
 | replicaCount                                       | int | `1` |  |
+| runtimeClassName                                   | string | `""` | RuntimeClassName refers to a RuntimeClass object in the node.k8s.io group, which should be used to run this pod. If unset or empty, the "legacy" RuntimeClass will be used. |
```
KasparMetsa: Could we maintain parameter table being alphabetically sorted. It should be fixed if you run `helm-docs -c packaging/helm-charts`

KasparMetsa: Could we add "If no RuntimeClass resource matches the named class, the pod will not be run." Running `helm-docs -c packaging/helm-charts` should fix it?

KasparMetsa (changes requested):

packaging/helm-charts/ksml/templates/statefulset.yaml
xsreality: Updated.

packaging/helm-charts/ksml/templates/job.yaml
xsreality: Updated.

packaging/helm-charts/ksml/README.md
xsreality: `helm-docs` is changing the README completely so manually edited for now.

xsreality: `helm-docs` is changing the README completely so manually edited for now.

KasparMetsa (approved):

PR #570, opened 2026-05-20
Silent integer overflow error

pantelis17:
# Fix silent integer overflow in Avro schema evolution

## The bug

A producer writes an Avro record with a `long` field. Later, the schema is
changed so the same field is `int`. A consumer reads the old record with the
new schema. KSML converts the value without checking the range. A value of
`3000000010` becomes `-1294967286`. No error. The bad record continues
through the pipeline.

## The fix

A new helper called `NumericRangeChecker` checks the range before every
narrowing cast (`long` to `int`, `double` to `float`, and so on). If the
value does not fit, it throws `DataException`. The pipeline stops with a
clear message:

```
Value 3000000001 exceeds INT range [-2147483648, 2147483647]
```

The check is used in three places that did a narrowing cast before:

- Avro write path: `long`, `float`, `double` to Avro `INT`, `LONG`, `FLOAT` fields.
- Python list to bytes conversion.
- Kafka header value (list of numbers to `byte[]`).

## Other small fixes

While checking the conversion paths, more silent failures were found. Each
one was a case where bad data did not throw an error before. They are fixed
in the same PR.

**Avro UNION branch selection** — always picked the first non-null branch, so `DataLong` in `[DOUBLE, LONG]` landed in `DOUBLE`. Avro then rejected it with a confusing internal `ClassCastException`. Now picks the branch whose native type matches first; no match → `DataException` at the conversion site.

**Avro ENUM validation** — an undeclared symbol or a non-String/DataEnum value silently yielded `null`, producing a generic "cannot be null" error from the Avro library. Now validates the symbol against the schema and throws with the allowed list.

**Avro type mismatches** — `BOOLEAN`, `BYTES`, `FIXED`, `STRING`, `ARRAY`, `MAP`, and `RECORD` fields fell through to a `fromDataObject()` fallback on wrong-type input, which produced values the Avro library rejected later with opaque errors. Now throws `DataException` at the conversion site naming the actual and expected types.

**Avro numeric widening** — `LONG`, `DOUBLE`, and `FLOAT` fields only accepted the exact matching KSML type; e.g. a `DataInteger` into a `LONG` field also hit the `fromDataObject()` fallback. Now explicitly accepts any numerically compatible type (narrowing paths are range-checked).

**SOAP body scalar values** — `DataInteger`, `DataLong`, `DataFloat`, `DataDouble`, and `DataBoolean` were silently dropped from the SOAP body because the serialiser used independent `if` branches instead of `if/else-if`. Now scalars are stringified, `DataNull` leaves the element empty, and unknown types throw.

**Protobuf `oneOf` field** — if no branch matched, the field was silently dropped from the message. Now throws `DataException` naming the unmatched value type and field name.

**Protobuf field type error** — a type mismatch (e.g. sending a `Long` to a `FLOAT` field) surfaced as a generic `ClassCastException` from deep inside the Protobuf library. Now caught and rethrown as `DataException` naming the field and offending type.

**`ConvertUtil` and `NativeDataObjectMapper` narrowing casts** — all narrowing conversions (e.g. `Long → int`, `Double → short`, `Float → byte`) called Java's `.intValue()` / `.shortValue()` directly, silently truncating out-of-range values. Now all paths go through `NumericRangeChecker` and throw on overflow.

**`ConvertUtil.convertStringToDataObject`** — `parseOrNull` swallowed all `NumberFormatException`s and always returned `null`, ignoring the `allowFail` flag. A malformed string (e.g. `"abc"` for an `int` field) silently disappeared instead of failing. Renamed to `parseOrFail`; now throws `DataException` when `allowFail=false`.

**`DefinitionParser` integer/long fields** — oversized YAML literals (e.g. `9999999999` for an `int` field) were silently narrowed by Jackson (e.g. to `1410065407`). Now validated with `node.isInt()` / `node.isLong()` and rejected with a `ParseException` naming the field and actual value.

**`JsonNodeUtil` JSON parse failure** — parse exceptions were silently swallowed (`// Ignore`). Now logged at DEBUG level so malformed JSON in pipelines is diagnosable.

## Tests

- New: `NumericRangeCheckerTest`, `DefinitionParserFieldValidationTest`
- New mapper tests for `HeaderDataObjectMapper`, `SoapDataObjectMapper`, `ProtobufDataObjectMapper` (the latter two modules had no mapper tests before)
- Existing Avro, Native, Convert, and Python tests extended for the new range checks
- End-to-end docker-compose test: KSML 1.2.1 reproduces the bug; this branch fixes it

## Not in scope

Avro logical types (`uuid`, `time-millis`, `time-micros`, `date`, `timestamp-*`, `decimal`) are passed through without semantic validation. Tracked by an inline TODO in `AvroDataObjectMapper`.


pantelis17:
[TESTS_EXECUTED.md](https://github.com/user-attachments/files/28104155/TESTS_EXECUTED.md)

pantelis17:
Added files with all the KSML pipelines which executed during testing phase

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
 
     private Object convertDataObjectToAvroBySchema(DataObject value, Schema schema) {
```
tonvanbart, line 315: This method now becomes very long and complex (nested `case` statements with nested conditionals inside, we might consider breaking it up or extracting some shorter methods?
I also see some same or almost-same logic and magic numbers (`0x1.0p63f `), suggest constants with an explanatory name

KasparMetsa:
Hey, regarding CSV. These PR comments are outdated:
```
CRITICAL — CSV required field with an empty cell silently dropped the field; now throws.
CRITICAL — CSV row shorter than schema silently padded; now throws.
```
The PR reverted these changes and just added these code comments:
image

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
...
         return switch (schema.getType()) {
             case NULL -> null;
-            case BOOLEAN -> value instanceof DataBoolean val ? val.value() : fromDataObject(value);
+            case BOOLEAN -> value instanceof DataBoolean val ? val.value() : schemaMismatch(value, schema);
             case INT -> switch (value) {
                 case DataByte val -> val.value() != null ? val.value().intValue() : null;
                 case DataShort val -> val.value() != null ? val.value().intValue() : null;
                 case DataInteger val -> val.value();
-                default -> fromDataObject(value);
+                case DataLong val -> {
```
KasparMetsa: `NumericRangeChecker.requireIntRange(long)` performs exactly the same `< Integer.MIN_VALUE || > Integer.MAX_VALUE` check and throws DataException with a message that already includes the bounds. The only piece of information lost is the trailing hint `"Use 'long' type in schema or ensure values fit in INT range."` - that hint should be moved into `requireIntRange` itself so all callers of the helper see it.

Suggested change:
```
case DataLong val -> {
    if (val.value() == null) yield null;
    NumericRangeChecker.requireIntRange(val.value());
    yield val.value().intValue();
}
```

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
...
+                    if (val.value() == null) yield null;
+                    long longVal = val.value();
+                    if (longVal < Integer.MIN_VALUE || longVal > Integer.MAX_VALUE) {
+                        throw new DataException(("Value %d exceeds INT range [%d, %d]. Use 'long' type in schema "
+                                + "or ensure values fit in INT range.")
+                                .formatted(longVal, Integer.MIN_VALUE, Integer.MAX_VALUE));
+                    }
+                    yield (int) longVal;
+                }
+                case DataFloat val -> {
```
KasparMetsa: `NumericRangeChecker.requireIntRange(double)` checks `!Double.isFinite(value) || value < Integer.MIN_VALUE || value > Integer.MAX_VALUE`. A `float` widens to a `double` losslessly - `Float.isFinite(f) == Double.isFinite((double) f)`, and the bounds inequalities are equivalent. The error message becomes the helper's shared one, which matches the inline message word-for-word ("Value %s cannot be converted to INT (out of range or not finite)").

**Suggested change:**

```java
case DataFloat val -> {
    if (val.value() == null) yield null;
    NumericRangeChecker.requireIntRange(val.value().doubleValue());
    yield (int) (float) val.value();
}
```

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
...
+                    if (val.value() == null) yield null;
+                    float floatVal = val.value();
+                    if (!Float.isFinite(floatVal) || floatVal < Integer.MIN_VALUE || floatVal > Integer.MAX_VALUE) {
+                        throw new DataException(
+                                "Value %s cannot be converted to INT (out of range or not finite)".formatted(floatVal));
+                    }
+                    yield (int) floatVal;
+                }
+                case DataDouble val -> {
```
KasparMetsa: `NumericRangeChecker.requireIntRange(double)` performs the identical finiteness + bounds check. `Double` auto-unboxes to `double` for the helper call; for the cast on the next line, `val.value().intValue()` is exactly `(int) (double) val.value()` because `Double.intValue()` is specified to truncate toward zero - identical to the inline `(int) doubleVal`.

**Suggested change:**

```java
case DataDouble val -> {
    if (val.value() == null) yield null;
    NumericRangeChecker.requireIntRange(val.value());
    yield val.value().intValue();
}
```

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
...
+                    yield (int) doubleVal;
+                }
+                default -> schemaMismatch(value, schema);
+            };
+            case LONG -> switch (value) {
+                case DataByte val -> val.value() != null ? val.value().longValue() : null;
+                case DataShort val -> val.value() != null ? val.value().longValue() : null;
+                case DataInteger val -> val.value() != null ? val.value().longValue() : null;
+                case DataLong val -> val.value();
+                case DataFloat val -> {
```
KasparMetsa: `NumericRangeChecker.requireLongRange(double)` already encodes the `>= 0x1.0p63 || < -0x1.0p63` boundary (with the comment "2^63 is exactly representable as a double; Long.MAX_VALUE = 2^63 - 1 is not"). Widening the `float` to `double` does not change the comparison - a float value at the `0x1.0p63f` boundary equals `0x1.0p63` as a double. The helper also performs the non-finite check up front. The thrown error message ("exceeds LONG range") gains the explicit numeric bounds the inline message left out.

**Suggested change:**

```java
case DataFloat val -> {
    if (val.value() == null) yield null;
    NumericRangeChecker.requireLongRange(val.value().doubleValue());
    yield (long) (float) val.value();
}
```

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
...
+                    if (val.value() == null) yield null;
+                    float floatVal = val.value();
+                    if (!Float.isFinite(floatVal) || floatVal >= 0x1.0p63f || floatVal < -0x1.0p63f) {
+                        throw new DataException(
+                                "Value %s cannot be converted to LONG (out of range or not finite)".formatted(floatVal));
+                    }
+                    yield (long) floatVal;
+                }
+                case DataDouble val -> {
```
KasparMetsa: The inline check and `requireLongRange(double)` are bit-identical - same finiteness check, same `0x1.0p63` boundary. `Double.longValue()` is specified to truncate toward zero, identical to `(long) doubleVal`.

**Suggested change:**

```java
case DataDouble val -> {
    if (val.value() == null) yield null;
    NumericRangeChecker.requireLongRange(val.value());
    yield val.value().longValue();
}
```

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
...
+            // long values above 2^24) but we deliberately do not reject that — see
+            // requireFloatRange's Javadoc for the same reasoning applied to DataDouble. If we
+            // rejected here, common pipeline values like DataInteger(1_000_000) → FLOAT would fail.
+            case FLOAT -> switch (value) {
+                case DataByte val -> val.value() != null ? val.value().floatValue() : null;
+                case DataShort val -> val.value() != null ? val.value().floatValue() : null;
+                case DataInteger val -> val.value() != null ? val.value().floatValue() : null;
+                case DataLong val -> val.value() != null ? val.value().floatValue() : null;
+                case DataFloat val -> val.value();
+                case DataDouble val -> {
```
KasparMetsa: The inline check is `Double.isFinite(value) && Math.abs(value) > Float.MAX_VALUE`. `NumericRangeChecker.requireFloatRange(double)` uses the identical predicate (see `NumericRangeChecker.java:147-150`). The intentional design - NaN and ±Infinity pass through because `(float)` preserves them - is preserved by the helper and is documented in the helper's Javadoc. `Double.floatValue()` is specified as `(float) doubleValue()`.

**Suggested change:**

```java
case DataDouble val -> {
    if (val.value() == null) yield null;
    NumericRangeChecker.requireFloatRange(val.value());
    yield val.value().floatValue();
}
```

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroDataObjectMapper.java
```diff
@@ -313,46 +313,229 @@ private Map<String, Object> convertDataStructToPlainMap(DataStruct struct) {
...
             case INT -> switch (value) {
                 case DataByte val -> val.value() != null ? val.value().intValue() : null;
                 case DataShort val -> val.value() != null ? val.value().intValue() : null;
                 case DataInteger val -> val.value();
-                default -> fromDataObject(value);
+                case DataLong val -> {
+                    if (val.value() == null) yield null;
+                    long longVal = val.value();
+                    if (longVal < Integer.MIN_VALUE || longVal > Integer.MAX_VALUE) {
+                        throw new DataException(("Value %d exceeds INT range [%d, %d]. Use 'long' type in schema "
```
KasparMetsa: This is helpful message. Currently when user hits similar in `NumericRangeChecker.java:93`, they don't get the nice explanation that user gets here.

### Suggested change

Move the hint into `NumericRangeChecker.requireIntRange` (and add equivalents on `requireLongRange`, `requireByteRange`, `requireShortRange`):

```java
public static void requireIntRange(long value) {
    if (value < Integer.MIN_VALUE || value > Integer.MAX_VALUE) {
        throw new DataException(
                ("Value %d exceeds INT range [%d, %d]. Use 'long' type in schema "
                + "or ensure values fit in INT range.")
                .formatted(value, Integer.MIN_VALUE, Integer.MAX_VALUE));
    }
}
``

File: ksml/src/main/java/io/axual/ksml/parser/DefinitionParser.java [hunk 9]
```diff
@@ -195,10 +195,29 @@ protected StructsParser<String> enumField(String childName, EnumSchema schema) {
         return new FieldParser<>(childName, false, schema.doc(), enumParser);
     }
 
+    /**
+     * Builds a parser for an integer-valued YAML field.
+     *
+     * <p>TODO: silently narrows oversized YAML literals. {@link ParseNode#asInt()} calls Jackson's
```
KasparMetsa: The goal of the Issue was to remove "silent integer overflow" yet the fix leaves two more silent narrowings in place, with only these Javadoc TODOs. Same situation for `longField`.

The entry point - parsing user YAML - still silently overflowing. A YAML literal 9999999999 becoming 1410065407 is the same class of bug as 3000000010 becoming -1294967286.

### Suggested changes like these:

```java
public Integer get() {
    var node = parseNode.get(childName);
    if (node == null) return null;
    if (!node.isInt()) {
        throw new TopologyException(
                "YAML value for '" + childName + "' is out of INT range: " + node.asText());
    }
    return node.intValue();
}
```

ksml-data/src/main/java/io/axual/ksml/data/util/NumericRangeChecker.java
```diff
@@ -0,0 +1,152 @@
...
+     * precision when cast to float (e.g. {@code 0.1} becomes {@code 0.10000000149...}), because
+     * that would reject very common pipeline values and diverge from Java's standard {@code (float)}
+     * cast semantics used by other JVM serialization frameworks (Avro, Jackson, Protobuf).</p>
+     *
+     * @param value the candidate value
+     * @throws DataException if {@code value} is finite and its magnitude exceeds {@link Float#MAX_VALUE}
+     */
+    public static void requireFloatRange(double value) {
+        if (Double.isFinite(value) && Math.abs(value) > Float.MAX_VALUE) {
+            throw new DataException("Value %s exceeds FLOAT range".formatted(value));
```
KasparMetsa:
```suggestion
            throw new DataException("Value %s exceeds FLOAT range [%s, %s]".formatted(value, -Float.MAX_VALUE, Float.MAX_VALUE));
```

KasparMetsa (changes requested):

ksml-data/src/main/java/io/axual/ksml/data/util/NumericRangeChecker.java
```diff
@@ -0,0 +1,152 @@
...
+ * limitations under the License.
+ * =========================LICENSE_END==================================
+ */
+
+import io.axual.ksml.data.exception.DataException;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/**
+ * Internal range-check helpers for narrowing numeric conversions.
```
KasparMetsa: The PR adds 317 comment lines out of 742 total lines added (43%).
- Some comments start with "Previously this did X" or "The previous behaviour was Y". That belongs in the PR description.
- Several comments justify the choice against alternatives that may never have been considered.
- Two methods (`setMessageFieldValue` in protobuf, `addChildToElement` in SOAP) have a multi-paragraph Javadoc AND a multi-line inline comment at the top of the body that say the same thing.

Can we please reduce the number of comments, make the comments we want shorter and make them easy to understand?

KasparMetsa:
Can we please improve the PR description, currently hard to understand what exactly was fixed in "Side fixes (uncovered while auditing)".

Maybe something like this:
[pr-570-suggested-description.md](https://github.com/user-attachments/files/28154716/pr-570-suggested-description.md)

KasparMetsa (changes requested):

ksml-data/src/test/java/io/axual/ksml/data/util/NumericRangeCheckerTest.java
```diff
@@ -0,0 +1,211 @@
...
+        assertThatCode(() -> NumericRangeChecker.requireFloatRange(0.0)).doesNotThrowAnyException();
+        assertThatCode(() -> NumericRangeChecker.requireFloatRange(Float.MAX_VALUE)).doesNotThrowAnyException();
+        assertThatCode(() -> NumericRangeChecker.requireFloatRange(-Float.MAX_VALUE)).doesNotThrowAnyException();
+    }
+
+    // --- convert methods ---
+
+    @Test
+    void convertLongToByte_returnsCorrectValue() {
+        assert NumericRangeChecker.convertLongToByte(42L) == (byte) 42;
```
KasparMetsa: The new tests added for the convert methods (`convertLongToByte`, `convertLongToShort`, `convertLongToInt`, `convertDoubleToByte`, `convertDoubleToShort`, `convertDoubleToInt`, `convertDoubleToLong`, `convertDoubleToFloat`) use Java's `assert` keyword.

This is a problem because:
1. **`assert` is disabled by default at runtime.** Java strips assertion checks unless the JVM is started with `-ea` (enable assertions). Maven Surefire's `<enableAssertions>` defaults to `true`, so the tests will fire in `mvn test`, but any future invocation that forgets `-ea` (an IDE quick-run config, a different build script, a debug session) will silently report "passed" without actually evaluating the conditions. That defeats the point of the test.
2. **It is inconsistent with the rest of the file.** The pre-existing tests in the same file use `assertThatCode(...).doesNotThrowAnyException()` and `assertThatThrownBy(...)`. Mixed assertion libraries in one test file are a maintenance hazard - the next test author has to pick one of two patterns and is likely to copy the closer example.



Fix:

Use the same AssertJ form as the rest of the file:

```java
@Test
void convertLongToByte_returnsCorrectValue() {
    assertThat(NumericRangeChecker.convertLongToByte(42L)).isEqualTo((byte) 42);
    assertThat(NumericRangeChecker.convertLongToByte((long) Byte.MIN_VALUE)).isEqualTo(Byte.MIN_VALUE);
    assertThat(NumericRangeChecker.convertLongToByte((long) Byte.MAX_VALUE)).isEqualTo(Byte.MAX_VALUE);
}

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #573, opened 2026-05-21
Added extra safeguard against path traversal

tonvanbart:
Added fix for AXPD-11242: PythonContext now normalizes paths before checking, so any `../../` in the path is normalized away. Also added tests.


jeroenvandisseldorp (approved):

ksml/src/main/java/io/axual/ksml/python/PythonContext.java
```diff
@@ -195,12 +198,20 @@ private FileSystem createReadOnlyFileSystem(String modulePath) {
         log.debug("createReadOnlyFileSystem({})", modulePath);
         FileSystem modulesFileSystem = FileSystem.newDefaultFileSystem();
         FileSystem denyAllAccess = FileSystem.newDenyIOFileSystem();
-        String sysPrefix = getPythonSysPrefix();
+        // Pre-normalize the allowed prefixes once - the per-call selector then only has to
+        // normalize the candidate path and run two startsWith checks.
+        final Path modulePathNormalized = Path.of(modulePath).toAbsolutePath().normalize();
```
pantelis17, line 222: We can use `toRealPath()` instead. `normalize()` doesn't resolve symlinks, so a symlink under modulePath can still escape the sandbox.

ksml/src/main/java/io/axual/ksml/python/PythonContext.java
```diff
@@ -195,12 +198,20 @@ private FileSystem createReadOnlyFileSystem(String modulePath) {
...
+        final Path sysPrefixNormalized = Path.of(getPythonSysPrefix()).toAbsolutePath().normalize();
```
pantelis17, line 223: Same here

ksml/src/main/java/io/axual/ksml/python/PythonContext.java
```diff
@@ -195,12 +198,20 @@ private FileSystem createReadOnlyFileSystem(String modulePath) {
...
         FileSystem restricted = FileSystem.newCompositeFileSystem(
                 // the default/fallback file system is: deny access
                 denyAllAccess,
-                // add a selector that returns the modulesFileSystem only if the path matches modulePath or sys.prefix
-                FileSystem.Selector.of(modulesFileSystem, path -> path.startsWith(modulePath) || path.startsWith(sysPrefix))
+                // add a selector that returns the modulesFileSystem only if the normalized path
+                // is within modulePath or sys.prefix
+                FileSystem.Selector.of(modulesFileSystem, path -> {
+                    Path normalized = path.toAbsolutePath().normalize();
```
pantelis17, line 233: We could reject `null` and non-absolute paths at the top of the lambda. I think that `toAbsolutePath()` resolves relative paths against the JVM's current working directory, so the same input can pass or fail depending on how the process was started. For example launched from `/allowed/modules`, the relative input `secrets.txt` resolves to `/allowed/modules/secrets.txt` and passes while launched from `/`, it resolves to `/secrets.txt` and is denied. Requiring absolute paths makes the check deterministic. Not sure if we can actually start the Jar from another directory but i think that its good to be guarded for this case.

ksml/src/main/java/io/axual/ksml/python/PythonContext.java
```diff
@@ -195,12 +198,20 @@ private FileSystem createReadOnlyFileSystem(String modulePath) {
...
+                    return normalized.startsWith(modulePathNormalized)
```
pantelis17: We could log denied paths at warn level before returning `false`. Right now denials are silent, so a typo and an attacker probing with `/etc/passwd`, `/etc/shadow` etc look the same

pantelis17 (approved):

PR #592, opened 2026-06-08
Added schema evolution compatibility tests

pantelis17:
## Summary                                                                                                                    
                                                                                                                                
Adds unit test coverage for all 18 schema evolution scenarios across AVRO, JSON Schema, and Protobuf, and fixes two schema compatibility bugs discovered in the process.  

### Production fixes
                                                                                              
- `DataSchema` — numeric type assignability now uses widening-only hierarchy (`LONG` accepts `INT`, but `INT` does not accept `LONG`; `DOUBLE` accepts `FLOAT`, but not vice-versa). Previously all integer types accepted each other unconditionally.      
- `StructSchema` — `isAssignableFrom` now rejects a required field whose counterpart in the other schema is optional. This correctly flags optional→required as a backward-incompatible change.
- `JsonSchemaMapper` / `ProtobufFileElementSchemaMapper` — minor constructor simplifications using the field required-flag correctly.                                                
                                                                                                                                
### Test coverage added
                             
- `AvroSchemaEvolutionCompatibilityTest` — all 18 matrix scenarios                                                            
- `JsonSchemaEvolutionCompatibilityTest` — all 18 matrix scenarios (union tests disabled: not applicable for JSON Schema)
- `ProtobufSchemaEvolutionCompatibilityTest` — all 18 matrix scenarios                                                        
- Integration tests extended: Avro backward compat (add optional field + int→long promotion); JSON Schema backward compat (add optional field)                                                                                                              
                                                                                                                                
## Tests Done                                                                                                                  
                                                                                                                                
- [ ] `mvn test -pl ksml-data,ksml-data-avro,ksml-data-json,ksml-data-protobuf` —> all pass
- [ ] Integration tests (`AvroSchemaEvolutionBackwardCompatibilityIT`, `JsonSchemaEvolutionIT`) run with Docker available


ksml-integration-tests/src/test/java/io/axual/ksml/integration/JsonSchemaEvolutionIT.java
```diff
@@ -0,0 +1,223 @@
...
+ * KSML-internal schema checks are already covered by JsonSchemaEvolutionCompatibilityTest.
+ *
+ * <p>Scenarios tested here:
+ * <ol>
+ *   <li>Backward compat — add optional field (matrix scenario 1): new schema reads old data.</li>
+ * </ol>
+ */
+@Slf4j
+@Testcontainers
+class JsonSchemaEvolutionIT {
```
KasparMetsa: Could we please consider removing this IT test because:
1. IT tests are expensive, they take a long time in the pipeline, so we should only have them if they are valuable and help us test something we can't test with unit tests.
2. `JsonSchemaRegistryIT` already tests the full JSON Schema pipeline with Apicurio end-to-end: produce, process, consume, validate field structure. It uses `SensorDataTestUtil` for rich assertions.
3. `JsonSchemaEvolutionIT` tests the same stack (JSON Schema + Apicurio) but adds a phase1/phase2 schema swap. This would add unique value for Avro because Avro has binary type promotion that happens on the wire. JSON Schema has no binary wire format - it is plain JSON text. The "evolution" part of the scenario is already fully covered by the 18 scenarios in `JsonSchemaEvolutionCompatibilityTest` (unit tests). There is nothing in JSON Schema serialization that a unit test cannot reach.
4. It costs: A full Kafka + Apicurio startup (~30-50 seconds), 2 KSMLContainer start/stops, and Awaitility waits for 10 messages twice. Roughly 4 minutes of CI time to confirm: "10 JSON messages flow through a pipeline with a schema change," which `JsonSchemaRegistryIT` already confirms (without the schema change) and the unit tests confirm (the schema change logic).
5. Recommendation: delete `JsonSchemaEvolutionIT` and its resource folders (`json-schema-evolution-tests-phase1/`, `json-schema-evolution-tests-phase2/`)

ksml-integration-tests/src/test/java/io/axual/ksml/integration/AvroSchemaEvolutionBackwardCompatibilityIT.java
KasparMetsa, line 86: For this test, in phase 2, after deserializing the processed messages, assert that the new fields (`location`, `accuracy`) are present with their default null values. This is the actual proof that schema evolution worked: old data can be read with a new schema and missing fields get their defaults.

ksml-integration-tests/src/test/java/io/axual/ksml/integration/AvroSchemaEvolutionBackwardCompatibilityIT.java
```diff
@@ -190,29 +176,134 @@ private void runPhase2_ProcessWithSchemaV2() {
...
 
-    /**
-     * Gets the total message count in a topic by summing up the end offsets of all partitions.
-     *
-     * @param bootstrapServers Kafka bootstrap servers
-     * @param topicName        the topic to check
-     * @return total message count across all partitions
-     */
+    @Test
+    void testBackwardCompatibility_TypePromotion_IntToLong() {
```
KasparMetsa, line 211: In phase 2, deserialize and assert that the `reading` field is a `long` and its value matches what was produced as an `int`. Without this, the test cannot distinguish "type promotion worked" from "type promotion silently dropped the value."

ksml-integration-tests/src/test/java/io/axual/ksml/integration/AvroSchemaEvolutionBackwardCompatibilityIT.java
```diff
@@ -190,29 +176,134 @@ private void runPhase2_ProcessWithSchemaV2() {
...
+            consumerProps.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+            consumerProps.put(ConsumerConfig.GROUP_ID_CONFIG, "test-type-promo-phase1-verifier");
+            consumerProps.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+            consumerProps.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+
+            try (KafkaConsumer<String, String> consumer = new KafkaConsumer<>(consumerProps)) {
+                consumer.subscribe(Collections.singletonList("sensor_data_avro_type_promo_test"));
+                ConsumerRecords<String, String> records = KSMLRunnerTestUtil.pollWithRetry(consumer, Duration.ofSeconds(10));
+
+                assertThat(records.count()).as("Should have produced exactly 10 messages").isEqualTo(10);
```
KasparMetsa: I know this is in existing tests as well so could we change it for all IT-s please to make pipeline faster?

Schema evolution correctness is proven by 1 message. 10 messages only makes the Awaitility wait longer. Reduce to 3.

```java
// Current — waits up to 30 seconds for 10 messages, four times
.until(() -> getTopicMessageCount(...) >= 10)
assertThat(records.count()).isEqualTo(10);

// Suggested
.until(() -> getTopicMessageCount(...) >= 3)
assertThat(records.count()).isGreaterThanOrEqualTo(3);
```

ksml-integration-tests/src/test/java/io/axual/ksml/integration/AvroSchemaEvolutionBackwardCompatibilityIT.java
KasparMetsa, line 70: A third `@Test` method could be added to `AvroSchemaEvolutionBackwardCompatibilityIT` at no extra container cost (Kafka and Apicurio are already static `@Container` fields shared across all tests in the class):

```
testBackwardIncompatibility_RemoveRequiredField()
```

The scenario:

1. **Phase 1** — produce 3 messages with schema v1: `{ name: string, reading: int, unit: string }`. All three fields are required. Uses the same pattern as the existing phase 1 methods: start a `KSMLContainer` with a producer YAML and the v1 schema, wait for 3 messages, stop.

2. **Phase 2** — start a `KSMLContainer` configured with schema v2: `{ name: string, reading: int }` — `unit` has been removed. This is a backward-incompatible change: the new schema cannot read old data because old messages carry a `unit` field that the new schema no longer declares, and the old schema required `unit` which means old producers always wrote it.

3. **Assert** — the KSML container either fails to start, or starts and then stops with an error logged. The assertion is that the job does not silently process messages as if nothing went wrong. Specifically: the output topic receives zero messages, and the KSML container logs contain an error mentioning the field name or schema incompatibility.

This is the only failure-path scenario where an IT genuinely adds value over a unit test. The unit test in `AvroSchemaEvolutionCompatibilityTest` already proves that `StructSchema.isAssignableFrom()` returns `notAssignable` when a required field is removed. What that unit test cannot prove is how KSML behaves at runtime when the Avro deserializer encounters this — whether it throws a clear exception, logs a useful message, or silently drops messages. Only a full pipeline test with a real registry can answer that.

ksml-data/src/main/java/io/axual/ksml/data/schema/StructSchema.java
```diff
@@ -459,6 +460,10 @@ public Assignable isAssignableFrom(DataSchema otherSchema) {
             final var thatField = that.field(field.name());
             // If the field exists in the other schema, then validate its compatibility
             if (thatField != null) {
+                // if this field is required but other provides it as optional, old producers may not have included it
+                if (field.defaultValue() == null && thatField.defaultValue() != null) {
```
KasparMetsa: ### What the code is trying to do

When KSML reads a message from Kafka, it checks whether the message's schema is compatible with the schema KSML expects. One of those checks is:

> "If I require a field, does the incoming data also guarantee that field is present?"

If KSML requires a field called `temperature` but the incoming data schema marks `temperature` as optional, then some old messages might not have `temperature` at all. That is a problem, so KSML should reject the combination.

### What the code actually does

A field in KSML has two separate pieces of information:

```
required     → true / false   (is this field mandatory?)
defaultValue → some value / null   (what to use if the field is absent)
```

The `required` flag directly answers the question "is this field mandatory?" Instead of reading `required`, the PR reads `defaultValue` and asks "does this field have a default value or not?":

```java
// PR in StructSchema.java — new check
if (field.defaultValue() == null && thatField.defaultValue() != null) {
    return Assignable.notAssignable("Field is required in this schema but optional in other schema");
}

// existing check in the same method, same pattern
if (field.defaultValue() == null && thatField == null) {
    return Assignable.notAssignable("Other schema does not contain required field");
}
```

The assumption is: *required fields have no default, optional fields do*.

### Why that assumption is fragile

Avro allows a field to be required AND have a default value at the same time. Nothing in the `Field` class prevents it either — the full constructor accepts any combination:

```java
new StructSchema.Field(
    "temperature",
    DataSchema.DOUBLE_SCHEMA,
    null,        // doc
    0,           // tag
    true,        // required = YES
    false,       // constant
    someDefault  // but also has a default value
);
```

If a field is created this way, `defaultValue()` is not null — so the code concludes "this field is optional" even though `required` is `true`. The compatibility check gives the wrong answer silently, with no error.

### The fix

```java
// StructSchema.isAssignableFrom — new check
if (field.required() && !thatField.required()) {
    return Assignable.notAssignable("Field is required in this schema but optional in other schema");
}

// existing check
if (field.required() && thatField == null) {
    return Assignable.notAssignable("Other schema does not contain required field");
}
```

`field.required()` is the direct answer to the direct question. It does not depend on what the default value happens to be.

KasparMetsa (changes requested):

ksml-integration-tests/src/test/java/io/axual/ksml/integration/AvroSchemaEvolutionBackwardCompatibilityIT.java
pantelis17: I spent some time digging into this and wanted to share what I found because the direction of the evolution matters a lot here and it's a bit subtle.
                                                                                                                                                                         
The scenario you described, producing with schema v1 (with the unit field) and then processing with schema v2 (without unit), is actually what Avro calls a backward-compatible change. When the new schema v2 reads old data written with v1, Avro's schema resolution simply ignores the extra unit field in the wire data because the reader schema doesn't care about it. So KSML processes all messages happily and we'd never be able to assert 0 messages.

What actually breaks things is the opposite direction: produce with v2 (unit field removed), then try to read with a legacy consumer that still has v1 (unit field required, non-nullable, no default). When Avro tries to deserialize, it can't find a value for unit and throws "null value for (non-nullable) string at SensorData.unit". Although, this is kinda already covered from unit test `removeRequiredField_forward_fail`.               

Am i missing something or should we just ignore this IT?

pantelis17: Decided against adding this test, does not add enough value for the development effort

ksml-data/src/main/java/io/axual/ksml/data/schema/StructSchema.java
```diff
@@ -177,7 +178,7 @@ public Field(String name, DataSchema schema, String doc, int tag, boolean requir
          * @param constant Whether the field is constant and unmodifiable.
          */
         public Field(String name, DataSchema schema, String doc, int tag, boolean required, boolean constant) {
-            this(name, schema, doc, tag, required, constant, null);
+            this(name, schema, doc, tag, required, constant, required ? null : DataNull.INSTANCE);
```
KasparMetsa: ## What a field knows

A KSML field (`StructSchema.Field`) holds two separate facts:

- `required` - must a message include this field?
- `defaultValue` - what value to use when the field is missing? (a value, or `null` for "no default")

These are not the same thing. A field can be optional with no default. A field can also be required with a default. So KSML must keep the two facts apart and never guess one from the other.

## The problem

The PR changed one constructor (one way to build a field). Now optional fields get a special value `DataNull.INSTANCE` instead of `null`:

```java
// PR - 6-part constructor
public Field(String name, DataSchema schema, String doc, int tag, boolean required, boolean constant) {
    this(name, schema, doc, tag, required, constant, required ? null : DataNull.INSTANCE);
}
```

But only this one constructor does it. The other constructors still store `null`. So the same optional field can come out two different ways:

```java
new Field("color", STRING, null, 0, false, false);        // -> defaultValue = DataNull.INSTANCE
new Field("color", STRING, null, 0, false, false, null);  // -> defaultValue = null
```

## Why it matters

The schema writers read `field.defaultValue()` to decide if they should write a default. Three of them handle the special value; one does not:

- `AvroSchemaMapper:355` - turns the special value into Avro's "null default".
- `ProtobufFileElementSchemaMapper:235` - skips the special value (`!= DataNull.INSTANCE`).
- `XmlSchemaMapper:357` - skips the special value (`!= DataNull.INSTANCE`).
- `NativeDataSchemaMapper:176` - checks only for `null`, so the special value slips through:

```java
// NativeDataSchemaMapper - DataNull.INSTANCE is not null, so this runs for optional fields
if (field.defaultValue() != null) encodeDefaultValue(result, field.defaultValue());
```

So an optional field built with the changed constructor writes an extra `default: null` into the native schema, while the same field built another way writes none. The result depends on which constructor was used. This is silent and hard to find. (This is the N2 finding.)

## A better fix

Set the rule in one place - the main constructor that all the others call - and change the 6-part one back to `null`:

```java
// main (8-part) constructor - set the value here, once
this.defaultValue = (defaultValue == null && !required) ? DataNull.INSTANCE : defaultValue;
```

```java
// 6-part constructor - back to null
this(name, schema, doc, tag, required, constant, null);
```

ksml-integration-tests/src/test/java/io/axual/ksml/integration/AvroSchemaEvolutionBackwardCompatibilityIT.java
```diff
@@ -164,55 +156,183 @@ private void runPhase2_ProcessWithSchemaV2() {
...
+
+                records.forEach(record -> {
+                    log.debug("  Type-promo Phase 2 processed message: key={}", record.key());
+                    assertThat(record.key()).as("Key should match sensor pattern")
+                            .matches("sensor[0-9]");
+
+                    GenericRecord value = (GenericRecord) record.value();
+                    Object readingObj = value.get("reading");
+                    assertThat(readingObj).as("reading field should be a Long after type promotion").isInstanceOf(Long.class);
+                    assertThat((Long) readingObj).as("reading value should be between 1 and 10").isBetween(1L, 10L);
```
KasparMetsa: This assertion only checks a loose range:

```java
// PR
assertThat((Long) readingObj).as("reading value should be between 1 and 10").isBetween(1L, 10L);
```

Suggested - assert the exact promoted value per record, using the deterministic key:

```java
int sensorIndex = Integer.parseInt(record.key().substring("sensor".length()));
long expectedReading = sensorIndex + 1L; // producer: key sensorN carries reading N+1
assertThat((Long) readingObj)
        .as("int reading must survive promotion to long unchanged")
        .isEqualTo(expectedReading);
```

Why: the point of the test is that `int -> long` promotion preserves the value. A range check passes even if promotion mangled a value (produced 3, read as 7 - both are "between 1 and 10").

KasparMetsa (changes requested):

ksml-data-json/src/main/java/io/axual/ksml/data/notation/json/JsonSchemaMapper.java
```diff
@@ -190,7 +190,8 @@ private List<StructSchema.Field> convertFields(DataStruct properties, Set<String
             var spec = entry.getValue();
             if (spec instanceof DataStruct specStruct) {
                 var doc = specStruct.getAsString(DESCRIPTION_NAME);
-                var field = new StructSchema.Field(name, convertType(specStruct, referenceResolver), doc != null ? doc.value() : null, NO_TAG, requiredProperties.contains(name));
+                boolean required = requiredProperties.contains(name);
+                var field = new StructSchema.Field(name, convertType(specStruct, referenceResolver), doc != null ? doc.value() : null, NO_TAG, required);
```
KasparMetsa: ## JSON Schema loses `default` in both directions

The semantic is documented on the field, but `JsonSchemaMapper` does not honor it.

**First, add a constant** (next to the other `*_NAME` constants, lines 64-74). There is no default constant today, and the JSON Schema keyword is `default`:
```java
private static final String DEFAULT_NAME = "default";
```

**Read** - in the `convertFields(...)` method (around lines 186-194).

Read - PR:
```java
var field = new StructSchema.Field(name, convertType(specStruct, referenceResolver),
        doc != null ? doc.value() : null, NO_TAG, required);
```
Read - suggested (carry the JSON `default` into the field):
```java
var defaultValue = specStruct.containsKey(DEFAULT_NAME)
        ? jsonDefaultToDataObject(specStruct.get(DEFAULT_NAME))   // JSON null -> DataNull.INSTANCE
        : null;                                                   // absent -> no default
var field = new StructSchema.Field(name, convertType(specStruct, referenceResolver),
        doc != null ? doc.value() : null, NO_TAG, required, false, defaultValue);
```

**Write** - in the `convertType(...)` method (signature at line 351). Today `defaultValue` is only used there in the `STRING` + `constant` branch (lines 358-362) to build a one-value `enum`; it never emits a `default`. Add this near the end of the method so it applies to every type:

Write - suggested (emit `default` for any field that has one):
```java
if (defaultValue != null) {                 // non-null means "this field has a default"
    target.put(DEFAULT_NAME, defaultValue);  // DataNull.INSTANCE already serializes to JSON null
}
```

(Note: `DataNull.INSTANCE` is a singleton - do not write `new DataNull()`. Putting `DataNull.INSTANCE` here produces JSON `"default": null`, which is the correct output for a field whose default is null.)

Why: without this, an Avro/Protobuf field with a default loses it when written to JSON Schema, and a JSON Schema `default` is dropped on read. For a schema-evolution feature this is exactly the information that should survive.

KasparMetsa (changes requested):

ksml-data-json/src/main/java/io/axual/ksml/data/notation/json/JsonSchemaMapper.java
```diff
@@ -404,5 +414,8 @@ private void convertType(DataSchema schema, boolean constant, DataObject default
             }
             target.put(ANY_OF_NAME, members);
         }
+        if (defaultValue != null) {
+            target.put(DEFAULT_NAME, defaultValue);
```
KasparMetsa, line 418: ## The problem

The base function definition has a `type` default of `"generic"`. But each specific function spec
limits `type` to its own single value with an `enum`. After the change, the generated spec now looks
like this (this exact shape appears 21 times in the file):

```json
"type": {
  "default": "generic",
  "description": "The type of the function",
  "enum": ["aggregator"]
}
```
- `enum: ["aggregator"]` says the only allowed value here is `aggregator`.
- `default: "generic"` says "if you do not set it, use `generic`".
- But `generic` is **not** in the allowed list. The default contradicts the `enum`.

## Why it matters

This file exists to help people write KSML in their editor. A recent commit was specifically about
making those autocomplete defaults **correct**. This change does the opposite in one place:

- The editor sees `default: "generic"` and pre-fills `type: generic` for you.
- The editor also sees `enum: ["aggregator"]`, so it immediately underlines `type: generic` as
  invalid.
- The result: the editor fills in a value and then says its own value is wrong. Confusing, and it
  pushes users toward an invalid definition.

It is not a crash and not strictly broken JSON Schema (most validators do not force `default` to be in
`enum`), but it is a real quality problem for the autocomplete experience.


## A simple before/after

Before the change, the spec had no default here, so the editor just offered `aggregator`:

```json
"type": { "description": "The type of the function", "enum": ["aggregator"] }
```

After the change, it pre-fills an invalid `generic`:

```json
"type": { "default": "generic", "description": "...", "enum": ["aggregator"] }
```

## The fix

The file to change is the converter:
`ksml-data-json/src/main/java/io/axual/ksml/data/notation/json/JsonSchemaMapper.java`, in the
`convertType(...)` method.

Today it writes the default unconditionally at the end of the method:

```java
// current code
if (defaultValue != null) {
    target.put(DEFAULT_NAME, defaultValue);
}
```

### Recommended fix: only write a default the field actually allows

If the field has an `enum`, the default must be one of the enum values. If it is not, do not write it.
This keeps the good behavior (defaults on normal fields, and defaults that are a valid enum choice)
and only drops the contradictory ones.

```java
// fixed code
if (defaultValue != null && defaultAllowedByEnum(target, defaultValue)) {
    target.put(DEFAULT_NAME, defaultValue);
}

/** A default is fine when the field has no enum, or when the default is one of the enum values. */
private static boolean defaultAllowedByEnum(DataStruct target, DataObject defaultValue) {
    if (!(target.get(ENUM_NAME) instanceof DataList allowedValues)) {
        return true;                                   // no enum -> any default is fine
    }
    for (final DataObject allowed : allowedValues) {   // enum present -> default must be in it
        if (allowed.equals(defaultValue)) return true;
    }
    return false;                                      // default contradicts the enum -> skip it
}
```

For the broken case this turns:

```json
"type": { "default": "generic", "description": "...", "enum": ["aggregator"] }
```

back into:

```json
"type": { "description": "The type of the function", "enum": ["aggregator"] }
```

while a normal field like `country` (no enum, default `"US"`) still keeps its `default`, and a field
whose default really is one of its enum values still keeps it too.

One small detail to confirm when implementing: compare by value. The enum holds `DataString` items and
the default is usually a `DataString`, so `allowed.equals(defaultValue)` works. If a non-string default
is ever compared against a string enum, fall back to comparing their string forms.

### Alternative fix: fix the source instead

Rather than guarding in the converter, you can fix where the schema is built: when a function's `type`
is locked to a specific value, do not leave the base `generic` default on it (clear it, or set it to
the specific type). Then the converter has nothing wrong to emit. This is cleaner in theory but harder
to locate (it lives in the function definition parsers), so the converter guard above is the smaller,
safer change.

### Important: do not remove the default feature

Writing out `default` for real message schemas (Avro, JSON, Protobuf) is correct and wanted - that was
the point of the original change. The fix must only stop the **contradictory** default; it must keep
writing valid ones.

KasparMetsa (changes requested):

KasparMetsa (approved):

Issue #395, opened 2025-10-29
Fix/Accept Sonar issues


KasparMetsa:
https://sonarcloud.io/summary/new_code?id=ksml&branch=main

Currently there are 481 issues in Sonar for main branch. Need to review them and either accept or fix them.

pantelis17:
As a next step to this issue, we need to:
- Review and fix the remaining issues
- Add more unit tests to cover as more as possible the code. By doing that, we will also achieve to pass the sonarqube gate for the 60% coverage in the changed code.

PR #595, opened 2026-06-09
Fix SonarQube Issues

pantelis17:
Fixed trivial sonarqube reports. Partially resolved issue #395


ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroSchemaMapper.java
```diff
@@ -428,8 +428,8 @@ private Object toJsonShapedDefault(DataObject value) {
         // native mapper unwrap the KSML wrapper to its underlying Java value
         // (DataString -> String, DataInteger -> Integer, etc.). If unwrapping
         // produces a raw Java null, swap it for Avro's null sentinel.
-        final var native_ = NATIVE_MAPPER.fromDataObject(value);
-        return native_ == null ? JsonProperties.NULL_VALUE : native_;
+        final var nativeDataObject = NATIVE_MAPPER.fromDataObject(value);
```
jeroenvandisseldorp: A nativeDataObject is a contradiction in itself, like black snow. Since we convert from DataObject to native here, I'd call this variable "native", or simply "value"

ksml-data-csv/src/main/java/io/axual/ksml/data/notation/csv/CsvDataObjectConverter.java
```diff
@@ -36,22 +36,13 @@ public class CsvDataObjectConverter implements Notation.Converter {
 
     @Override
     public DataObject convert(DataObject value, DataType targetType) {
-        // Convert from structured CSV
-        if (value instanceof DataList valueList && DEFAULT_TYPE.equals(valueList.type())) {
-            // Convert to String
-            if (targetType == DataString.DATATYPE) {
-                return new DataString(MAPPER.fromDataObject(value));
-            }
+        if (value instanceof DataList valueList && DEFAULT_TYPE.equals(valueList.type()) && targetType == DataString.DATATYPE) {
```
jeroenvandisseldorp: For some conversions, it makes sense to keep nested ifs, since it makes it easier for developers to follow the reasoning. When you compress them into one statement with a lot of conditions, that logic is much, much harder to understand. I think here that is the case.

ksml-data-csv/src/main/java/io/axual/ksml/data/notation/csv/CsvDataObjectConverter.java
```diff
@@ -36,22 +36,13 @@ public class CsvDataObjectConverter implements Notation.Converter {
...
 
-        // Convert from String
-        if (value instanceof DataString csvString) {
-            // Convert to structured CSV
-            if (targetType instanceof ListType
-                    || targetType instanceof StructType
-                    || targetType instanceof UnionType) {
-                return MAPPER.toDataObject(targetType, csvString.value());
-            }
+        if (value instanceof DataString csvString &&
```
jeroenvandisseldorp: For some conversions, it makes sense to keep nested ifs, since it makes it easier for developers to follow the reasoning. When you compress them into one statement with a lot of conditions, that logic is much, much harder to understand. I think here that is the case.

ksml-data-protobuf-apicurio/src/main/java/io/axual/ksml/data/notation/protobuf/apicurio/ApicurioProtobufFileElementDescriptorMapper.java
```diff
@@ -124,7 +118,12 @@ private MessageDefinition toMessageDefinition(String namespace, MessageElement m
             for (final var field : messageElement.getFields()) {
                 final var required = field.getLabel() == null || field.getLabel() == Field.Label.REQUIRED;
                 final var repeated = field.getLabel() == Field.Label.REPEATED;
-                final var label = required ? null : repeated ? "repeated" : "optional";
+                final String label;
```
jeroenvandisseldorp: These single line statements are okay IMO. Writing them out to multiline gives them more screen space than they should have

ksml-data-protobuf-apicurio/src/main/java/io/axual/ksml/data/notation/protobuf/apicurio/ApicurioProtobufFileElementDescriptorMapper.java
```diff
@@ -222,9 +226,15 @@ private MessageElement convertToMessageElement(Descriptors.Descriptor messageDes
...
+            final Field.Label fieldLabel;
+            if (required) {
+                fieldLabel = null;
+            } else {
+                fieldLabel = list ? Field.Label.REPEATED : Field.Label.OPTIONAL;
+            }
             return new FieldElement(
                     DEFAULT_LOCATION,
-                    required ? null : list ? Field.Label.REPEATED : Field.Label.OPTIONAL,
+                    fieldLabel,
```
jeroenvandisseldorp: Same as above, more lines does not make this clearer or cleaner imo

ksml-data-protobuf-confluent/src/main/java/io/axual/ksml/data/notation/protobuf/confluent/ConfluentProtobufFileElementDescriptorMapper.java
```diff
@@ -127,7 +121,12 @@ private MessageDefinition toMessageDefinition(String namespace, MessageElement m
             for (final var field : messageElement.getFields()) {
                 final var required = field.getLabel() == null || field.getLabel() == Field.Label.REQUIRED;
                 final var repeated = field.getLabel() == Field.Label.REPEATED;
-                final var label = required ? null : repeated ? "repeated" : "optional";
+                final String label;
```
jeroenvandisseldorp: Same as above

ksml-data-protobuf-confluent/src/main/java/io/axual/ksml/data/notation/protobuf/confluent/ConfluentProtobufFileElementDescriptorMapper.java
```diff
@@ -182,7 +181,12 @@ private MessageElement convertToMessageElement(Descriptors.Descriptor messageDes
             for (final var oneOf : messageDescriptor.getOneofs()) {
                 final var oneOfFields = new ArrayList<FieldElement>();
                 for (final var oneOfField : oneOf.getFields()) {
-                    final var label = oneOfField.isRequired() ? Field.Label.REQUIRED : oneOfField.isRepeated() ? Field.Label.REPEATED : Field.Label.OPTIONAL;
+                    final Field.Label label;
```
jeroenvandisseldorp, line 184: Same as above

ksml-data-protobuf-confluent/src/main/java/io/axual/ksml/data/notation/protobuf/confluent/ConfluentProtobufFileElementDescriptorMapper.java
```diff
@@ -226,9 +230,15 @@ private MessageElement convertToMessageElement(Descriptors.Descriptor messageDes
...
+            final Field.Label fieldLabel;
+            if (required) {
+                fieldLabel = null;
+            } else {
+                fieldLabel = list ? Field.Label.REPEATED : Field.Label.OPTIONAL;
+            }
             return new FieldElement(
                     DEFAULT_LOCATION,
-                    required ? null : list ? Field.Label.REPEATED : Field.Label.OPTIONAL,
+                    fieldLabel,
```
jeroenvandisseldorp: Same as above

ksml-data-protobuf/src/main/java/io/axual/ksml/data/notation/protobuf/ProtobufFileElementSchemaMapper.java
```diff
@@ -228,9 +227,15 @@ private static FieldElement convertStructFieldToFieldElement(StructSchema.Field
...
+        final Field.Label fieldLabel;
+        if (required) {
+            fieldLabel = Field.Label.REQUIRED;
+        } else {
+            fieldLabel = list ? Field.Label.REPEATED : Field.Label.OPTIONAL;
+        }
         return new FieldElement(
                 DEFAULT_LOCATION,
-                required ? Field.Label.REQUIRED : list ? Field.Label.REPEATED : Field.Label.OPTIONAL,
+                fieldLabel,
```
jeroenvandisseldorp: Same as above

ksml-data/src/main/java/io/axual/ksml/data/compare/Equality.java
```diff
@@ -35,6 +36,7 @@
  * To optimize for the common success case, the OK state is implemented as a singleton instance (message is {@code null}).
  */
 @Getter
+@EqualsAndHashCode
```
jeroenvandisseldorp: Why do we need this?

ksml-data/src/main/java/io/axual/ksml/data/compare/Equality.java
```diff
@@ -56,7 +58,7 @@ private Equality(String message, Equality cause) {
      *
      * @return the OK Equal instance
      */
-    public static Equality equal() {
+    public static Equality equalResult() {
```
jeroenvandisseldorp: I get that this rename makes it less likely to confuse the method with equals(), but if we're concerned about that, let's reconsider the naming entirely instead of just this method

ksml-data/src/main/java/io/axual/ksml/data/compare/Equality.java
```diff
@@ -66,8 +68,8 @@ public static Equality equal() {
      * @param message explanation of why the objects are not equal
      * @return a new Equal instance representing inequality
      */
-    public static Equality notEqual(String message) {
-        return notEqual(message, null);
+    public static Equality notEqualResult(String message) {
```
jeroenvandisseldorp: I get that this rename makes it less likely to confuse the method with equals(), but if we're concerned about that, let's reconsider the naming entirely instead of just this method

ksml-data/src/main/java/io/axual/ksml/data/object/DataStruct.java
jeroenvandisseldorp, line 88: I think now that SOAP is removed, we can get rid of the DataStructApplier entirely....

ksml-data/src/main/java/io/axual/ksml/data/object/DataStruct.java
```diff
@@ -184,11 +183,7 @@ public DataObject get(String key) {
     public <T> void getIfPresent(String key, Class<T> clazz, DataStructApplier<T> applier) {
```
jeroenvandisseldorp, line 184: I think with SOAP removed, we can get rid of this method...

ksml-data/src/main/java/io/axual/ksml/data/compare/Equality.java
KasparMetsa: ### `Equality.equal()` / `Equality.notEqual()` renamed to `Equality.equalResult()` / `Equality.notEqualResult()`

**Sonar rule:** `S2176` - Method name should not shadow `Object.equals()`.

**Problem:** These are static factory methods on a custom `Equality` class - they don't override `Object.equals()`. The name `equal()` is a legitimate, well-understood factory method name (see `Optional.empty()`, `Comparator.naturalOrder()`). Renaming to `equalResult()` and `notEqualResult()` is verbose, awkward, and hurts readability across dozens of call sites.

This rename cascades across ~30 files and every `equals()` implementation in the domain model.

**PR change (before):**
```java
public static Equality equal() { return EQUAL; }
public static Equality notEqual(String message) { ... }
```

**PR change (after):**
```java
public static Equality equalResult() { return EQUAL; }
public static Equality notEqualResult(String message) { ... }
```

**Recommended fix instead:** Suppress at the class level in `Equality.java`:
```java
@SuppressWarnings("java:S2176")
public class Equality { ... }
```

This keeps the original clean names and avoids a 30-file rename that adds no value.

ksml-runner/src/main/java/io/axual/ksml/runner/producer/ExecutableProducer.java
```diff
@@ -95,7 +96,13 @@ private ExecutableProducer(UserFunction generator,
         this.valueSerializer = valueSerializer;
     }
 
-    private record GeneratedMessage(Headers headers, byte[] key, byte[] value) {
+    // Use class instead of record to prevent Sonar warning about equals() and hashCode() overrides
```
KasparMetsa: ### `GeneratedMessage` record converted to `@AllArgsConstructor @Getter` class

**Sonar rule:** `S6218` - `equals()` and `hashCode()` should be overridden in classes containing arrays.

**PR change (before):**
```java
private record GeneratedMessage(Headers headers, byte[] key, byte[] value) {
}
```

**PR change (after):**
```java
// Use class instead of record to prevent Sonar warning about equals() and hashCode() overrides
@AllArgsConstructor
@Getter
private static class GeneratedMessage {
    private Headers headers;
    private byte[] key;
    private byte[] value;
}
```

**Problem:** The record was semantically correct. `GeneratedMessage` is a pure data carrier with no equality semantics needed - it is never compared with `equals()`. Converting to a mutable class with Lombok introduces:
- `private` fields that are now mutable (no `final`)
- Loss of the structural `toString()` that records provide for free
- A misleading class (records communicate "this is immutable data")

The rule is correct that `byte[]` equality in records uses identity (`Arrays.equals` is not called), but in this case that does not matter - the object is never compared.

**Recommended fix instead:** Suppress at the record declaration:
```java
@SuppressWarnings("java:S6218")
private record GeneratedMessage(Headers headers, byte[] key, byte[] value) {
}
```

ksml-data/src/main/java/io/axual/ksml/data/util/ValuePrinter.java
```diff
@@ -42,7 +42,7 @@ public String print(Object value, boolean quoted) {
         final var quote = quoted && (value instanceof String || value instanceof DataString) ? dict.quoteStr() : "";
         return switch (value) {
             case null -> dict.nullStr();
-            case Boolean val -> val ? dict.trueStr() : dict.falseStr();
+            case Boolean val -> Boolean.TRUE.equals(val) ? dict.trueStr() : dict.falseStr();
```
KasparMetsa: **Sonar rule:** `S5411` - Boxed value should not be compared using `==` / unboxing concern.

**PR change (before):**
```java
case Boolean val -> val ? dict.trueStr() : dict.falseStr();
```

**PR change (after):**
```java
case Boolean val -> Boolean.TRUE.equals(val) ? dict.trueStr() : dict.falseStr();
```

**Problem:** Inside `case Boolean val ->`, the variable `val` is guaranteed to be a non-null `Boolean` by the pattern match itself - the JVM verifies this before entering the arm. Unboxing `val` in a ternary is completely safe here. `Boolean.TRUE.equals(val)` is unnecessary, less idiomatic, and slower (it's a method call with a null-check that can never trigger).

**Recommended fix instead:** Suppress the specific line, or add a blanket project-level exclusion for `S5411` inside switch pattern arms.

ksml-data/src/test/java/io/axual/ksml/data/serde/StringSerdeTest.java
```diff
@@ -70,7 +70,9 @@ void serializerThrowsForWrongValue() {
         var nativeMapper = new NativeDataObjectMapper();
         var serde = new StringSerde(nativeMapper);
 
-        assertThatThrownBy(() -> serde.serializer().serialize(TOPIC, new Exception("test")))
+        var serializer = serde.serializer();
```
KasparMetsa: **Sonar rule:** `S2095` / `S1941` - Variables referenced in lambdas should be effectively final (or local variable capture).

Examples throughout the test suite:

**PR change (before):**
```java
assertThatThrownBy(() -> serde.serializer().serialize(TOPIC, new Exception("test")))
```

**PR change (after):**
```java
var serializer = serde.serializer();
var wrongInput = new Exception("test");
assertThatThrownBy(() -> serializer.serialize(TOPIC, wrongInput))
```

This pattern repeats in `StringSerdeTest`, `UnionSerdeTest`, `DataObjectSerdeTest`, `NullSerdeTest`. The original one-liners were more readable and idiomatic for AssertJ. The extracted variables add noise that makes the test intent harder to read at a glance.

**Recommended fix instead:** Configure `sonar.issue.ignore.multicriteria` or add `@SuppressWarnings("java:S2095")` on test classes, since test lambdas capturing expressions are a standard pattern.

ksml-data-protobuf-apicurio/src/main/java/io/axual/ksml/data/notation/protobuf/apicurio/ApicurioProtobufFileElementDescriptorMapper.java
```diff
@@ -102,13 +98,11 @@ private MessageDefinition toMessageDefinition(String namespace, MessageElement m
...
             final var msgBuilder = MessageDefinition.newBuilder(messageElement.getName());
             for (final var nestedType : messageElement.getNestedTypes()) {
-                if (notDuplicateType(fullName, nestedType)) {
-                    if (nestedType instanceof MessageElement nestedMessage) {
-                        msgBuilder.addMessageDefinition(toMessageDefinition(fullName, nestedMessage));
-                    }
-                    if (nestedType instanceof EnumElement nestedEnum) {
-                        msgBuilder.addEnumDefinition(toEnumDefinition(fullName, nestedEnum));
-                    }
+                if (notDuplicateType(fullName, nestedType) && nestedType instanceof MessageElement nestedMessage) {
```
KasparMetsa: **PR change (before):**
```java
if (notDuplicateType(fullName, nestedType)) {
    if (nestedType instanceof MessageElement nestedMessage) {
        msgBuilder.addMessageDefinition(toMessageDefinition(fullName, nestedMessage));
    }
    if (nestedType instanceof EnumElement nestedEnum) {
        msgBuilder.addEnumDefinition(toEnumDefinition(fullName, nestedEnum));
    }
}
```

**PR change (after):**
```java
if (notDuplicateType(fullName, nestedType) && nestedType instanceof MessageElement nestedMessage) {
    msgBuilder.addMessageDefinition(toMessageDefinition(fullName, nestedMessage));
}
if (notDuplicateType(fullName, nestedType) && nestedType instanceof EnumElement nestedEnum) {
    msgBuilder.addEnumDefinition(toEnumDefinition(fullName, nestedEnum));
}
```

**Problem:** `notDuplicateType()` is now called twice per nested type (once for the `MessageElement` check, once for the `EnumElement` check). If the method has any side effects (e.g., it registers types into a set), calling it twice would corrupt that state. Even if it's currently pure, this is an unintended functional change that violates the original intent.

**Fix:** Keep the outer guard:
```java
if (notDuplicateType(fullName, nestedType)) {
    if (nestedType instanceof MessageElement nestedMessage) { ... }
    if (nestedType instanceof EnumElement nestedEnum) { ... }
}
```

ksml-data/src/main/java/io/axual/ksml/data/object/DataStruct.java
```diff
@@ -90,9 +90,8 @@ public interface DataStructApplier<T> {
          * Applies the operation to the specified value.
          *
          * @param value the value to be processed.
-         * @throws Exception if the operation fails.
          */
-        void apply(T value) throws Exception;
+        void apply(T value);
```
KasparMetsa: **PR change (before):**
```java
public interface DataStructApplier<T> {
    /**
     * @throws Exception if the operation fails.
     */
    void apply(T value) throws Exception;
}
```

**PR change (after):**
```java
public interface DataStructApplier<T> {
    void apply(T value);
}
```

And in `getIfPresent()`:
```java
// Before:
try {
    applier.apply((T) value);
} catch (Exception e) {
    throw new DataException("Exception thrown while getting and applying DataStruct value for key \"" + key + "\"", e);
}

// After:
applier.apply((T) value);
```

**Problem:** This is a breaking API change. Any existing lambda or anonymous class that throws a checked exception from `apply()` will no longer compile. The original `try/catch` was intentional defensive code - it wrapped any exception thrown by an applier into a `DataException` with context about which key caused the failure. That context is now permanently lost. If an applier ever throws a `RuntimeException`, it will now propagate unchecked and uncontextualized.

ksml-integration-tests/src/test/java/io/axual/ksml/integration/testutil/KSMLRunnerTestUtil.java
```diff
@@ -466,7 +467,7 @@ public static void waitForSchemaRegistryReady(String registryUrl, Duration timeo
                 return;
             }
 
-            Thread.sleep(2000); // Check every 2 seconds
+            Utils.sleep(2000); // Check every 2 seconds
```
KasparMetsa: **PR change (before):**
```java
Thread.sleep(2000);
Thread.sleep(1000);
```

**PR change (after):**
```java
Utils.sleep(2000); // org.apache.kafka.common.utils.Utils
Utils.sleep(1000);
```

**Problem:** `org.apache.kafka.common.utils.Utils.sleep()` swallows `InterruptedException` silently - it catches it and calls `Thread.currentThread().interrupt()` but does not rethrow. In integration tests this hides test termination signals and can cause tests to hang or produce incorrect results when interrupted. `Thread.sleep()` is the correct choice in test code because it forces the caller to handle interruption explicitly.

ksml-data/src/test/java/io/axual/ksml/data/serde/NullSerdeTest.java
```diff
@@ -43,12 +43,16 @@ void serializerAlwaysReturnsNull() {
     @Test
     @DisplayName("deserializer returns null for null/empty bytes; throws for non-empty")
     void deserializerReturnsNullOrThrows() {
-        var serde = new NullSerde();
-        assertThat(serde.deserializer().deserialize(TOPIC, null)).isNull();
-        assertThat(serde.deserializer().deserialize(TOPIC, new byte[]{})).isNull();
+        try (var serde = new NullSerde()) {
```
KasparMetsa, line 47: **PR change (before):**
```java
var serde = new NullSerde();
assertThat(serde.deserializer().deserialize(TOPIC, null)).isNull();
// ...
assertThatThrownBy(() -> serde.deserializer().deserialize(TOPIC, new byte[]{1}))
        .isInstanceOf(DataException.class);
```

**PR change (after):**
```java
try (var serde = new NullSerde()) {
    assertThat(serde.deserializer().deserialize(TOPIC, null)).isNull();
    // ...
    var deserializer = serde.deserializer();
    assertThatThrownBy(() -> deserializer.deserialize(TOPIC, new byte[]{1}))
            .isInstanceOf(DataException.class);
} catch (Exception _) {
    // ignore
}
```

**Problem:** The outer `catch (Exception _) { // ignore }` silently swallows any assertion failure or unexpected exception from the test body. If a test assertion fails inside the `try` block, the failure will be caught and discarded - the test will appear to pass when it should fail. This is a serious test reliability issue.

If the intent was only to handle `close()` failure, the correct pattern is:
```java
try (var serde = new NullSerde()) {
    // assertions here - no outer catch
}
```
`AutoCloseable.close()` exceptions from try-with-resources are handled by the JVM and do not need a catch block.

KasparMetsa (changes requested):

ksml-data/src/main/java/io/axual/ksml/data/util/ConvertUtil.java
```diff
@@ -431,12 +431,10 @@ public static DataObject convertNullToDataObject(DataType expected) {
         if (expected == DataBytes.DATATYPE) return new DataBytes();
         if (expected == DataString.DATATYPE) return new DataString();
         return switch (expected) {
-            case EnumType ignored -> DataNull.INSTANCE;
             case ListType listType -> new DataList(listType.valueType(), true);
             case MapType mapType -> new DataMap(mapType.valueType(), true);
             case StructType structType -> new DataStruct(structType.schema(), true);
-            case TupleType ignored -> DataNull.INSTANCE;
-            case UnionType ignored -> DataNull.INSTANCE;
+            case EnumType _, TupleType _, UnionType _ -> DataNull.INSTANCE;
```
jeroenvandisseldorp: While this is logically still correct, I prefer having cases on their own line for understandability

ksml-data/src/test/java/io/axual/ksml/data/serde/NullSerdeTest.java
```diff
@@ -43,12 +43,16 @@ void serializerAlwaysReturnsNull() {
...
+            assertThat(serde.deserializer().deserialize(TOPIC, null)).isNull();
+            assertThat(serde.deserializer().deserialize(TOPIC, new byte[]{})).isNull();
 
-        assertThatThrownBy(() -> serde.deserializer().deserialize(TOPIC, new byte[]{1}))
-                .isInstanceOf(DataException.class)
-                .hasMessageEndingWith("Can only deserialize empty byte arrays as DataNull");
+            var deserializer = serde.deserializer();
```
jeroenvandisseldorp: For all cases: prefer "final var" everywhere, even in test classes

ksml-integration-tests/src/test/java/io/axual/ksml/integration/ApicurioAvroSchemaRegistryIT.java
```diff
@@ -102,9 +102,9 @@ void testKSMLApicurioAvroToJsonConversion() throws Exception {
             log.info("Found {} Apicurio AVRO sensor messages", records.count());
 
             // Log some sample sensor data keys (values will be binary AVRO)
-            records.forEach(record -> {
-                log.info("🔬 Apicurio AVRO Sensor: key={}, value size={} bytes", record.key(), record.value().length());
-                assertThat(record.key()).as("Sensor key should start with 'sensor'").startsWith("sensor");
+            records.forEach(consumerRecord -> {
```
jeroenvandisseldorp, line 105: What is the logic here? Records is a list, so why not name an element the same singular?

ksml-runner/src/main/java/io/axual/ksml/runner/KSMLRunner.java
```diff
@@ -483,7 +484,7 @@ private static void printRunnerSchema(String filename) {
                 writer.close();
                 log.info("KSML JSON schema written to file: {}", filename);
             } else {
-                System.out.println(schema);
+                log.info("{}", schema);
```
jeroenvandisseldorp: This is not right, as it changes the output. We want a clean line with just the JSON schema here, not something that gets prefixed with date/time and a log level.

ksml-runner/src/test/java/io/axual/ksml/runner/logging/KSMLLogbackConfiguratorTest.java
```diff
@@ -107,7 +109,7 @@ void configureWithoutEnvironmentVariable() {
         KSMLLogbackConfigurator configurator = new KSMLLogbackConfigurator();
         configurator.setContext(spiedContext);
         configurator.configure(spiedContext);
-        System.out.println(MockAppender.APPENDERS);
+        log.info("{}", MockAppender.APPENDERS);
```
jeroenvandisseldorp: Not sure if this is okay or changes the desired output

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/AssertionRunner.java
```diff
@@ -222,15 +213,32 @@ private static Object toNativeValue(Object value) {
         return value;
     }
 
+    private void injectVariables(PythonContext pythonContext, List<Pair<String, Object>> args) {
+        for (var nameValue : args) {
+            var code = """
```
jeroenvandisseldorp, line 218: final var

ksml-test-runner/src/test/java/io/axual/ksml/testrunner/TestDataProducerTest.java
```diff
@@ -193,7 +194,8 @@ void undefinedStreamReferenceThrows() {
                 List.of(new TestMessage("k1", "v1", null)),
                 null, null);
 
+        var blocks2 = List.of(block);
```
jeroenvandisseldorp: blockList?

jeroenvandisseldorp (changes requested):

ksml-data-avro/src/main/java/io/axual/ksml/data/notation/avro/AvroSchemaMapper.java
pantelis17: It cant be native as it is reserved word, also it cant be value as we already have it defined. I will rename it to "object"

ksml-data/src/main/java/io/axual/ksml/data/object/DataStruct.java
pantelis17: We can open a separate issue for this to avoid more changes

ksml-data/src/main/java/io/axual/ksml/data/util/ConvertUtil.java
pantelis17: Honestly, i prefer grouping them and find it quite understandable. If you strongly disagree i will split it

ksml-integration-tests/src/test/java/io/axual/ksml/integration/ApicurioAvroSchemaRegistryIT.java
pantelis17: record is a reserved keyword in Java. Its a lot better to use something else instead of suppressing the error

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/AssertionRunner.java
```diff
@@ -170,13 +161,13 @@ private List<Map<String, Object>> collectOutputRecords(StreamDefinition stream)
 
         var records = new ArrayList<Map<String, Object>>();
         var keyValues = outputTopic.readRecordsToList();
-        for (var record : keyValues) {
+        for (var keyValue : keyValues) {
```
pantelis17: fixed, i also prefer final var everywhere  so we can have claude pass the whole project and replace named types with var and add final wherever is possible. Let me know if you want to open this task

ksml-data/src/main/java/io/axual/ksml/data/object/DataTuple.java
```diff
@@ -125,12 +125,10 @@ public Equality equals(Object other, EqualityFlags flags) {
...
 
         // Compare contents
-        if (!flags.isSet(IGNORE_DATA_TUPLE_CONTENTS)) {
-            if (elements() != null || that.elements() != null) {
-                if (elements() == null || that.elements() == null) return EqualUtil.objectNotEqual(this, that);
-                final var contentsEqual = equalContents(this, that, flags);
-                if (contentsEqual.isNotEqual()) return objectNotEqual(this, that, contentsEqual);
-            }
+        if (!flags.isSet(IGNORE_DATA_TUPLE_CONTENTS) && (elements() != null || that.elements() != null)) {
+            if (elements() == null || that.elements() == null) return EqualUtil.objectNotEqual(this, that);
```
tonvanbart, line 129: In line 128, did the confition not verify that both `elements()` and `that.elements()` are both not null? Can this condition ever happen?

jeroenvandisseldorp (approved):

PR #623, opened 2026-06-18
Fix handling of nested schema in apicurio serdes

### LINKED ISSUE #290: Schema with nested Schema is not working with Apicurio SerDes when `apicurio.registry.auto-register: false`
We have a Schema `SensorData` that has a nested Schema `SensorType`, the schema below is uploaded to the Platform Manager

```
{
  "type": "record",
  "name": "SensorData",
  "namespace": "io.axual.ksml.example",
  "doc": "Emulated sensor data with a few additional attributes",
  "fields": [
    {
      "name": "name",
      "type": "string",
      "doc": "The name of the sensor"
    },
    {
      "name": "timestamp",
      "type": "long",
      "doc": "The timestamp of the sensor reading"
    },
    {
      "name": "value",
      "type": "string",
      "doc": "The value of the sensor, represented as string"
    },
    {
      "name": "type",
      "type": {
        "type": "enum",
        "name": "SensorType",
        "doc": "The type of a sensor",
        "symbols": [
          "AREA",
          "HUMIDITY",
          "LENGTH",
          "STATE",
          "TEMPERATURE"
        ]
      },
      "doc": "The type of the sensor"
    },
    {
      "name": "unit",
      "type": "string",
      "doc": "The unit of the sensor"
    },
    {
      "name": "color",
      "type": [
        "null",
        "string"
      ],
      "doc": "The color of the sensor",
      "default": null
    },
    {
      "name": "city",
      "type": [
        "null",
        "string"
      ],
      "doc": "The city of the sensor",
      "default": null
    },
    {
      "name": "owner",
      "type": [
        "null",
        "string"
      ],
      "doc": "The owner of the sensor",
      "default": null
    }
  ]
}
```

When starting a KSML Application with Apicurio SerDes configured like these

```
# Specify all properties to be passed into Apicurio
apicurio.registry.auto-register: false
apicurio.registry.auto-register.if-exists: "RETURN"
```

we noticed that the KSML Application is trying to `register` in the Apicurio Registry a Schema that has a reference to the inner Schema.

```
{
    "type": "record",
    "name": "SensorData",
    "namespace": "io.axual.ksml.example",
    "doc": "Emulated sensor data with a few additional attributes",
    "fields": [
        {
            "name": "name",
            "type": "string",
            "doc": "The name of the sensor"
        },
        {
            "name": "timestamp",
            "type": "long",
            "doc": "The timestamp of the sensor reading"
        },
        {
            "name": "value",
            "type": "string",
            "doc": "The value of the sensor, represented as string"
        },
        {
            "name": "type",
            "type": "SensorType",
            "doc": "The type of the sensor"
        },
        {
            "name": "unit",
            "type": "string",
            "doc": "The unit of the sensor"
        },
        {
            "name": "color",
            "type": [
                "null",
                "string"
            ],
            "doc": "The color of the sensor",
            "default": null
        },
        {
            "name": "city",
            "type": [
                "null",
                "string"
            ],
            "doc": "The city of the sensor",
            "default": null
        },
        {
            "name": "owner",
            "type": [
                "null",
                "string"
            ],
            "doc": "The owner of the sensor",
            "default": null
        }
    ]
}
```

This is the log reported by the Apicurio Registry

```
apicurio.audit action="request" result="failure" src_ip="10.1.48.224" path="/apis/registry/v2/groups/default/artifacts/apicurio_sensordata-value/meta" response_code="404" method="POST" user=""
```

tonvanbart:
Fixes #290 

The issue is caused by SerdeConfig.FIND_LATEST_ARTIFACT defaulting to `false`. This fails for the nested schema because Apicurio computes a key for content based lookup in this case, but it computes the key as only a bare reference instead of as a full schema. Since the original schema has the full enum inline so the content IDs do not match.
The fix defaults the configuration for the Apicurio serde to `FIND_LATEST_ARTIFACT = true`. This means that the schema lookup now is done by getting the latest registered version for the subject.
My gut feeling is that this is what we want in 99% of cases, in cases where it is not wanted the configuration can be overridden. 
The PR adds a targeted IT test to prevent regression.


ksml-integration-tests/src/test/java/io/axual/ksml/integration/Issue290ApicurioNestedEnumNoAutoRegisterIT.java
```diff
@@ -0,0 +1,198 @@
...
+ */
+@Slf4j
+@Testcontainers
+class Issue290ApicurioNestedEnumNoAutoRegisterIT {
+
+    private static final String TOPIC = "sensor_data_avro_290";
+    private static final String ARTIFACT_ID = TOPIC + "-value";
+    private static final String GROUP = "default";
+
+    static final Network network = Network.newNetwork();
```
KasparMetsa, line 86: `Network.newNetwork()` creates a Docker network, but `network.close()` is never called. The network stays in Docker after the test finishes. Every test run makes a new one and leaves it behind. On a CI server that runs the tests again and again, these dead networks keep piling up.

**Suggested fix:**

```java
static final Network network = Network.newNetwork();

@AfterAll
static void teardown() {
    network.close();
}
```

Now the network is removed at the end of the test class, so nothing is left behind.

ksml-integration-tests/src/test/java/io/axual/ksml/integration/Issue290ApicurioNestedEnumNoAutoRegisterIT.java
```diff
@@ -0,0 +1,198 @@
...
+ * artifactId {@code <topic>-value} in group {@code default}, then runs a KSML producer
+ * that serializes {@code avro:SensorData} to that topic with auto-register disabled, and
+ * checks whether messages actually land on the topic.
+ *
+ * <p><b>Status: passes, guarding the fix.</b> Without the fix this test FAILS: with the
+ * issue-faithful config (only {@code auto-register: false} + {@code auto-register.if-exists:
+ * RETURN}) Apicurio's content-based lookup represents the nested enum as a bare reference,
+ * which never matches the inline-registered schema, so serialization fails with
+ * {@code ArtifactNotFoundException}. The fix in {@code ApicurioAvroSerdeSupplier} defaults
+ * {@code apicurio.registry.find-latest=true}, so the serializer resolves the artifact's
```
KasparMetsa, line 66: ## What the PR changed

The PR sets the lookup strategy to latest-version by default:

- Before: the default was **content-based** (`find-latest = false`).
- After: the default is **latest-version** (`find-latest = true`).

## Suggested fix

1. **Add a CHANGELOG entry** for the release, saying that the Apicurio Avro default for
   `apicurio.registry.find-latest` changed from `false` to `true`, and why (issue #290).
2. **Add a note in the Apicurio notation config reference** explaining the new default and how to go
   back to the old behavior, for example:

   ```yaml
   notations:
     avro:
       type: apicurio_avro
       config:
         apicurio.registry.find-latest: false   # restore the old content-based lookup
   ```

Now an upgrading user can see the change, understand it, and opt out if they need the old behavior.

KasparMetsa (changes requested):

ksml-integration-tests/src/test/java/io/axual/ksml/integration/Issue290ApicurioNestedEnumNoAutoRegisterIT.java
```diff
@@ -0,0 +1,205 @@
...
+@Testcontainers
+class Issue290ApicurioNestedEnumNoAutoRegisterIT {
+
+    private static final String TOPIC = "sensor_data_avro_290";
+    private static final String ARTIFACT_ID = TOPIC + "-value";
+    private static final String GROUP = "default";
+
+    static final Network network = Network.newNetwork();
+
+    static void teardown() {
```
KasparMetsa, line 89: The `@AfterAll` annotation is missing. JUnit will never call this method automatically. The leak is still present.

// what was added:
```java
static void teardown() {   // <- @AfterAll is missing
    network.close();
    schemaRegistry.stop();
    kafka.stop();
    ksml.stop();
}
```

// what it should be:
```java
@AfterAll
static void teardown() {
    network.close();
}
```

Note: the container `.stop()` calls for schemaRegistry, kafka, and ksml are also unnecessary - those are annotated with `@Container`, so Testcontainers handles them already. The only thing needed is network.close() under `@AfterAll`.

docs/reference/configuration-reference.md
```diff
@@ -305,6 +305,18 @@ ksml:
...
 ```
 
+!!! note
+    As of release 1.3.0 Apicurio AVRO has the default for `apicurio.registry.find-latest` changed to `true` (was: `false`).
+    The previous behavior is obtained by overriding the new default as follows:
+    ```yaml
+        notations:
+          avro:
+            type: apicurio_avro
+              config:
```
KasparMetsa: Both a release-notes.md entry and a configuration-reference.md note were added - that is correct. However, the YAML example in the note has wrong indentation. `config`: is indented as a child of type:, but it should be a sibling of type::

what is in the docs (wrong - config is a child of type):
```
    notations:
      avro:
        type: apicurio_avro
          config:                                   # <- 2 extra spaces too many
            apicurio.registry.find-latest: false
```

what it should be (matching the existing examples on line 280-284):
```
    notations:
      avro:
        type: apicurio_avro
        config:                                     # <- same level as type
          apicurio.registry.find-latest: false
```

This will result in a YAML parse error.

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #625, opened 2026-06-19
Fix incorrect schema resolving flags in UserTypeParser calling code

jeroenvandisseldorp:
This PR solves a bug where certain points in the code called the wrong userType method in DefinitionParser, effectively preventing key/valueType fields in topic definitions from being parsed as `unresolved`. This PR removes the userType method with `allowUnresolved=false` entirely and reroutes its callers to the proper one with explicit allowUnresolved parameter. Callers should set this to `true` in parsers that parse topic definitions, and false everywhere else.


ksml/src/main/java/io/axual/ksml/definition/parser/KeyValueStateStoreDefinitionParser.java
```diff
@@ -44,8 +44,8 @@ protected StructsParser<KeyValueStateStoreDefinition> parser() {
...
         final var segmentIntervalField = optional(durationField(KSMLDSL.Stores.SEGMENT_INTERVAL, "Size of segments for storing old record versions (must be positive). Old record versions for the same key in a single segment are stored (updated and accessed) together. The only impact of this parameter is performance. If segments are large and a workload results in many record versions for the same key being collected in a single segment, performance may degrade as a result. On the other hand, historical reads (which access older segments) and out-of-order writes may slow down if there are too many segments."));
-        final var keyTypeField = optional(userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the keyValue store"));
-        final var valueTypeField = optional(userTypeField(KSMLDSL.Stores.VALUE_TYPE, "The value type of the keyValue store"));
+        final var keyTypeField = optional(userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the keyValue store", true));
```
KasparMetsa: This comment is about **state stores**. It does not change the rest of PR (streams, tables and topics), which are correct.

## The problem

This PR set state store `keyType` / `valueType` to allow an "unresolved" type - a registry-backed notation written **without** a schema name (for example `valueType: avro` instead of `valueType: avro:SensorData`).

For a stream this is fine: a stream has a real input topic, so KSML fetches the schema from the registry using the topic name. A state store has **no input topic** to look the schema up by, and there is no step that resolves it. So the unresolved type slips through parsing and only blows up later, during topology build, with a message that does not say what is actually wrong.

## Reproduced (Docker, local image from this branch)

A keyValue store declared with `valueType: avro` (no schema name), used by an `aggregate`:

```yaml
stores:
  sensor_store:
    type: keyValue
    keyType: string
    valueType: avro            # registry notation, NO schema name
```

### Before (current PR branch) - confusing, late error

KSML starts, builds the topology, and then fails:

```
Description: KSML execution error: Incompatible value types for state store 'sensor_store':
             avro:Unresolved and avro:SensorData
  io.axual.ksml.operation.BaseOperation::validateStoreTypeWithStreamType @ BaseOperation.java:548
  io.axual.ksml.operation.AggregateOperation::apply @ AggregateOperation.java:81
  io.axual.ksml.TopologyGenerator::generate ...
```

Why this is bad:

- it says "Incompatible value types", which sounds like a type mismatch, not a missing schema name
- it appears only at topology build time, deep inside the aggregate operation
- it does not tell the user what to fix

### After (with the change below) - clear, early error

KSML now rejects it during parsing, before any topology is built:

```
Description: KSML parse error: Error in input node processor->stores->sensor_store:
             Unspecified schema can only be used in the context of a topic: avro:Unresolved
  io.axual.ksml.parser.DefinitionParser::lambda$userTypeField$0 @ DefinitionParser.java:238
```

Why this is better:

- it points at the exact place: `processor->stores->sensor_store`
- it explains the real cause: a schema name is needed here
- it fails fast, at parse time, before Kafka Streams starts

## The change required

### Code in the PR

State store key and value types are parsed with `allowUnresolved = true` (the last argument):

```java
// KeyValueStateStoreDefinitionParser
final var keyTypeField = optional(userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the keyValue store", true));
final var valueTypeField = optional(userTypeField(KSMLDSL.Stores.VALUE_TYPE, "The value type of the keyValue store", true));

// SessionStateStoreDefinitionParser
final var keyTypeField = userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the session store", true);
final var valueTypeField = userTypeField(KSMLDSL.Stores.VALUE_TYPE, "The value type of the session store", true);

// WindowStateStoreDefinitionParser
final var keyTypeField = userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the window store", true);
final var valueTypeField = userTypeField(KSMLDSL.Stores.VALUE_TYPE, "The value type of the window store", true);
```

### Code suggested

Change the last argument to `false` in all three parsers:

```java
// KeyValueStateStoreDefinitionParser
final var keyTypeField = optional(userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the keyValue store", false));
final var valueTypeField = optional(userTypeField(KSMLDSL.Stores.VALUE_TYPE, "The value type of the keyValue store", false));

// SessionStateStoreDefinitionParser
final var keyTypeField = userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the session store", false);
final var valueTypeField = userTypeField(KSMLDSL.Stores.VALUE_TYPE, "The value type of the session store", false);

// WindowStateStoreDefinitionParser
final var keyTypeField = userTypeField(KSMLDSL.Stores.KEY_TYPE, "The key type of the window store", false);
final var valueTypeField = userTypeField(KSMLDSL.Stores.VALUE_TYPE, "The value type of the window store", false);
```

### Why each change (explained simply)

The last argument is `allowUnresolved`. `true` means "a type with no schema name is allowed here; the schema will be fetched from the registry later". That only works when there is an input topic to look the schema up by. A state store has no input topic, so there is nothing to look it up by. So for stores the value must be `false`: the type has to be known at parse time.

- **keyValue store** (used by `count`, `reduce`, `aggregate`): set to `false`. The store has no topic to fetch a schema for, so a missing schema name is caught right away during parsing, not later in the aggregate step.
- **session store** (used by windowed aggregations): set to `false`, for the same reason - no topic, so the type must be given in full.
- **window store** (used by windowed aggregations): set to `false`, same reason.

In short: `true` only makes sense for things backed by a real topic (streams, tables, producer topics). State stores are not, so `false` is the correct value and it produces a clear, early error.

This does not break any valid setup:

- a store with a full type (`valueType: avro:SensorData`) still works
- a store with no `valueType` at all still works (the type comes from the operation)
- only the invalid case (`valueType: avro`, no schema name) is now rejected early with a clear message

## Tests

`DefinitionParserUnresolvedTypeTest` parses through the real parsers, so it locks the behavior in:

- stream value type without a schema name -> accepted (a stream may resolve from the registry)
- state store value type without a schema name -> rejected with the clear message
- function `resultType`, function parameter `type`, convert `into` without a schema name -> rejected

## Optional follow-up (nice to have, not required)

The parse-time message could be even friendlier by adding a hint, like the one already used elsewhere
in `UserTypeParser`:

```
Unspecified schema can only be used in the context of a topic. Use "avro:SchemaName" to set the schema.
```

That is a small wording change in `UserTypeParser` and would help all three "must be known" places (state stores, function types, convert targets). It is optional; the main fix above is the important part.

KasparMetsa (changes requested):

KasparMetsa (approved):

Issue #369, opened 2025-10-06
Add Schema Validation for ksml-runner.yaml


KasparMetsa:
## Why are we doing this?

Currently, KSML validates definition files (streams, pipelines, functions) using an auto-generated JSON schema (`ksml-language-spec.json`), but **ksml-runner.yaml configuration files are not validated**. This leads to configuration errors that are only discovered at deployment time, causing delayed feedback and harder debugging.

**Current problem:**
```yaml
notations:
  jsonschema:
    type: apicurio_jsonschema  # ✅ Valid
    schemaRegistry: my_apicurio_registry
    config: {}
```

vs

```yaml
notations:
  jsonschema:
    type: apicurio_json_schema  # ❌ Invalid - typo not caught until runtime
    schemaRegistry: my_apicurio_registry
    config: {}
```

The typo `apicurio_json_schema` (should be `apicurio_jsonschema`) is not detected until deployment, wasting time and causing confusion.

**Root cause:** The runner configuration classes (KSMLRunnerConfig, KSMLConfig, NotationConfig, etc.) use Jackson parsing without schema validation. These classes have `@JsonIgnoreProperties(ignoreUnknown = true)`, which silently ignores typos and invalid fields.

**Solution approach:** Generate a JSON schema for ksml-runner.yaml configuration using the same mechanism as KSML definitions:
- Create definition classes with definition parsers for runner configuration (similar to TopologyDefinitionParser for KSML definitions)
- Generate `ksml-runner-spec.json` schema alongside `ksml-language-spec.json`
- Validate ksml-runner.yaml files against the schema
- Provide clear error messages for invalid configurations

This enables:
1. **Early validation** - Catch syntax errors immediately, not at deployment
2. **IDE support** - Autocompletion and inline validation in IntelliJ/VS Code
3. **Clear errors** - Specific messages like "Invalid notation type: 'apicurio_json_schema'. Valid values are: apicurio_jsonschema, apicurio_avro, confluent_avro, apicurio_protobuf"
4. **Enum validation** - Catch invalid enum values (notation types, error handlers, etc.) at parse time

### 1. Schema Generation
- [ ] Create definition and parser classes for ksml-runner configuration (KSMLRunnerConfig, KSMLConfig, NotationConfig, etc.)
- [ ] Generate `ksml-runner-spec.json` JSON schema using the same JsonSchemaMapper approach as ksml-language-spec.json
- [ ] Schema covers all configuration sections:
[...]

PR #631, opened 2026-06-23
Add schema validation tests for ksml-runner.yaml examples

### LINKED ISSUE #393: Add test like AllDefinitionsSchemaValidationTest for ksml-runner.yaml examples
- `AllDefinitionsSchemaValidationTest` tests that all example KSML definitions pass the validation of `ksml-language-spec.json` schema

- We need a similar test for all ksml-runner.yaml examples to check if they pass the validation of `ksml-runner-spec.json` as was done in https://github.com/Axual/ksml/issues/369

We need to test ksml-runner.yaml files validation at least in these locations:
- /Users/km/dev/ksml/ksml/src/test/resources/docs-examples
- /Users/km/dev/ksml/ksml-integration-tests/src/test/resources/docs-examples
- /Users/km/dev/ksml/examples

There are problems with circular dependency currently. Here are ideas how to solve it: https://github.com/Axual/ksml/pull/391#discussion_r2470380752

- Also check that configuration reference doc matches the generated schema and Improve either the schema or the conf reference docs.

pantelis17:
Description
                                                                                                                                                                                
  ## Summary                                                                                                                                                                  
  - New `ksml-docs-test` module consolidates schema-validation tests for KSML definition                                                                                        
    files **and** `ksml-runner.yaml` files, validating every example against a freshly                                                                                        
    generated schema on each test run.        
  - Adds a `NotationType` enum so the runner schema actually rejects invalid notation
    type values like `confluentavro` or `apicurio_json_schema` (closes the enum                                                                                                 
    acceptance criterion from #369 that was left open by #391).                                                                                                                 
  - Fixes three example notations in `examples/ksml-runner.yaml` that produced YAML                                                                                             
    `null` under a `config:` key and tripped the new validation.                                                                                                                
                                                                                                                                                                              
  Closes #393. References #369.                                                                                                                                                 
                                                                                                                                                                                
  ## Background                                                                                                                                                                 
                                                                                                                                                                                
  Issue #393 asked for a test mirroring `AllDefinitionsSchemaValidationTest`, but for                                                                                         
  `ksml-runner.yaml` files in:
  - `ksml/src/test/resources/docs-examples`   
  - `ksml-integration-tests/src/test/resources/docs-examples`
  - `examples/`                               
                                                                                                                                                                                
  The blocker called out in #391's review was the circular dependency between `ksml`
  (where the definitions test lived) and `ksml-runner` (where the runner schema is                                                                                              
  generated). The cleanest resolution suggested there was a dedicated module for docs                                                                                         
  testing — that's what this PR introduces.                                                                                                                                     
                                                                                                                                                                              
  ## Changes                                                                                                                                                                    
                                                                                                                                                                                
  ### New module: `ksml-docs-test`            
  Single home for both schema validation suites. Depends on `ksml` and `ksml-runner`                                                                                            
  (test scope) — no edges going back, so no circular dependency.                                                                                                              
                                                                                                                                                                                
  - `AllDefinitionsSchemaValidationTest` — moved here from `ksml`. Generates the
    language schema in-memory via `TopologyDefinitionParser` + `JsonSchemaMapper`,                                                                                              
    walks every `*.yaml` (except `ksml-runner.yaml`) under                                                                                                                    
    `ksml/src/test/resources` and `ksml-integration-tests/src/test/resources/docs-examples`.                                                                                    
    Runs 282 parameterized cases.                                                                                                                                             
  - `AllRunnerConfigSchemaValidationTest` — new. Regenerates the runner schema by                                                                                               
    calling `KSMLRunner.main("--runner-schema", tempFile)` (same code path the Maven                                                                                            
    build uses to produce the committed `docs/ksml-runner-spec.json`), walks every                                                                                              
    `ksml-runner.yaml` across the three required directories. Runs 16 parameterized                                                                                             
    cases.                                                                                                                                                                      
                                                                                                                                                                              
  Because both schemas are produced from current source on every test run, there is                                                                                             
  no "stale committed schema" failure mode regardless of how tests are invoked                                                                                                  
  (`mvn test -pl ksml-docs-test`, full reactor build, or IDE).                                                                                                                  
                                                                                                                                                                                
  ### `NotationType` enum in `NotationConfig`                                                                                                                                 
  Previously `type` was a free `String`, so typos like `apicurio_json_schema` (the                                                                                              
  literal example called out in #369) silently passed validation. Now it's a proper                                                                                             
  enum mirroring the `Handler` pattern, with 8 values matching what                                                                                                             
  `NotationFactories` accepts at runtime:                                                                                                                                       
                                                                                                                                                                                
  - Schema-registry-backed: `apicurio_avro`, `confluent_avro`, `apicurio_jsonschema`,                                                                                           
    `apicurio_protobuf`                                                                                                                                                         
  - Built-in: `csv`, `xml`, `json`, `binary`                                                                                                                                    
                                                                                                                                                                                
  VicTools now emits an `"enum": [...]` constraint in `NotationType` in
  `ksml-runner-spec.json`, and the validator reports a clear actionable error when                                                                                              
  the value isn't on the list:                                                                                                                                                  
                                                                                                                                                                                
  $.ksml.notations.avro.type: does not have a value in the enumeration                                                                                                          
    ["apicurio_avro", "confluent_avro", "apicurio_jsonschema", "apicurio_protobuf",                                                                                             
     "csv", "xml", "json", "binary"]                                                                                                                                            
                                          
  The only consumer of `NotationConfig.type()` — `KSMLRunner.java` — now calls                                                                                                  
  `.type().jsonValue()` to get the wire-name string for the factory lookup.                                                                                                     
                                              
  ### TODOs for unwired Confluent providers                                                                                                                                     
  `ConfluentJsonSchemaNotationProvider` and `ConfluentProtobufNotationProvider`                                                                                                 
  exist as classes but their `META-INF/services/...NotationProvider` files are                                                                                                  
  empty, so they aren't discovered by `ServiceLoader` and would fail at runtime.                                                                                                
  TODO comments have been added to both SPI files explaining what to uncomment if                                                                                               
  those providers are ever wired up, with a pointer to also extend `NotationType`.                                                                                              
                                              
  ### Example file fixes                                                                                                                                                        
  - `examples/ksml-runner.yaml`: three notations (`second_avro`, `jsonschema`,                                                                                                  
    `protobuf`) had `config:` with all values commented out, parsing as YAML null                                                                                               
    and failing the schema's "object expected" check. Commented out the `config:`                                                                                               
    key itself in each — semantically the same intent.                                                                                                                          
                                                                                                                                                                              
  ### Schema regeneration                                                                                                                                                       
  - `docs/ksml-runner-spec.json` regenerated to reflect the new `NotationType`                                                                                                  
    enum (the `process-classes` phase of `ksml-runner` produces this file).                                                                                                     
                                                                                                                                                                                
  ## Test plan                                                                                                                                                                  
  - [ ] `mvn test -pl ksml-docs-test --also-make -Dsonar.skip=true` — 298 cases green                                                                                         
  - [ ] `mvn test -pl ksml-runner -Dsonar.skip=true` — 42 cases green (incl. existing                                                                                           
        `KSMLRunnerTest`)                     
  - [ ] Confirm `docs/ksml-runner-spec.json` contains                                                                                                                           
        `"NotationType"` with all 8 enum values                                                                                                                                 
  - [ ] Negative-test sanity check: temporarily set a notation `type:` to a typo                                                                                                
        like `confluentavro` in any `ksml-runner.yaml` → confirm                                                                                                                
        `AllRunnerConfigSchemaValidationTest` fails with the enum violation message                                                                                             
        and a JSON-pointer path → revert and confirm green


jeroenvandisseldorp (approved):

ksml-runner/src/main/java/io/axual/ksml/runner/config/NotationConfig.java
```diff
@@ -35,11 +37,64 @@
...
+        public static NotationType forValue(String value) {
+            if (value == null) {
+                return null;
+            }
+            for (NotationType nt : values()) {
+                if (nt.jsonValue.equals(value)) {
+                    return nt;
+                }
+            }
+            return null;
```
KasparMetsa: ## What the code does

When KSML Runner starts, it reads a `ksml-runner.yaml` config file. This file contains notation entries like:

```yaml
notations:
  avro:
    type: confluent_avro
```

The `type` value tells the runner which serializer to use. The runner looks up that serializer by name and registers it. If the serializer is not found, the notation is broken and KSML cannot read or write data.

---

## The problem

### Before this PR

The `type` field was a plain `String`. When the runner looked up the serializer with that string and found nothing, it threw a `ConfigException`:

```
Unknown notation type: confluentavro
```

The runner stopped immediately at startup. The user saw a clear error message with the bad value.

### After this PR

The `type` field is now a `NotationType` enum. Jackson reads the string from YAML and calls `forValue()` to convert it to an enum constant. If the string is unknown, `forValue()` returns `null` instead of throwing an error:

```java
return null;  // unknown string -> no error thrown
```

The runner then checks `if (factoryName != null)` before registering the serializer. When `factoryName` is null, it skips the registration and only logs a warning:

```
Notation configuration incomplete: notation=avro, serde=null
```

The runner starts successfully with a broken notation. There is no error, no crash - just a warning that is easy to miss.

---

## Why this is a problem

Imagine a user writes a typo in their config:

```yaml
notations:
  avro:
    type: confluentavro   # wrong - should be confluent_avro
```

**Before this PR:** The runner crashes at startup with `Unknown notation type: confluentavro`. The user fixes the typo and tries again.

**After this PR:** The runner starts without any error. The `avro` notation is never registered. When KSML tries to process a message, it fails later with a confusing error about a missing notation - not about the typo in the config. The user has to trace back through logs to find the original warning.

The PR adds schema validation tests that check for this typo before deployment. But those tests only run when you run the test suite. The runtime behavior (what happens when the runner actually starts) is now more lenient than before.

---

## The fix

**Current PR code** (`NotationConfig.java`):

```java
@JsonCreator
public static NotationType forValue(String value) {
    if (value == null) {
        return null;
    }
    for (NotationType nt : values()) {
        if (nt.jsonValue.equals(value)) {
            return nt;
        }
    }
    return null;  // unknown string -> caller gets null, logs a warning, continues
}
```

**Suggested change** - throw instead of returning `null`:

```java
@JsonCreator
public static NotationType forValue(String value) {
    if (value == null) {
        return null;
    }
    for (NotationType nt : values()) {
        if (nt.jsonValue.equals(value)) {
            return nt;
        }
    }
    throw new IllegalArgumentException("Unknown notation type: " + value +
            ". Valid values: " + Arrays.stream(values())
                .map(NotationType::jsonValue)
                .collect(Collectors.joining(", ")));
}
```

Jackson will catch this and throw a `JsonMappingException` when it reads the config file. The runner stops immediately at startup with a clear message:

```
Unknown notation type: confluentavro. Valid values: apicurio_avro, confluent_avro, ...
```

This is the same behavior as before the PR, just with a better error message.

The `unknownTypeDeserializesAsNullType` test in `NotationTypeTest` should also be updated to expect this exception instead of asserting that `cfg.type()` is null.

ksml-runner/src/main/java/io/axual/ksml/runner/config/NotationConfig.java
```diff
@@ -35,11 +37,64 @@
...
         @JsonPropertyDescription("The name of the Schema Registry configuration to use, if applicable for the type")
         String schemaRegistry,
         @JsonProperty(value = "config", required = false)
         @JsonPropertyDescription("Additional properties for the serializer, these will be added to any configuration options defined for the the schema registry")
         StringMap config) {
+
+    @JsonClassDescription("Supported notation serializer implementations.")
+    public enum NotationType {
+        // Schema-registry-backed notations (loaded via ServiceLoader)
+        @JsonProperty("apicurio_avro")
```
KasparMetsa: **Why this matters:** Each constant carries both `@JsonProperty("apicurio_avro")` and the constructor argument `"apicurio_avro"`. The `@JsonProperty` annotations do nothing - `@JsonCreator forValue()` overrides deserialization and `@JsonValue jsonValue()` overrides serialization, so Jackson never reads the per-constant annotations. If someone fixes a `@JsonProperty` string but not the constructor arg (or vice versa), they diverge silently with no compile error.

**PR code:**

```java
@JsonProperty("apicurio_avro")
APICURIO_AVRO("apicurio_avro"),
@JsonProperty("confluent_avro")
CONFLUENT_AVRO("confluent_avro"),
// ... same pattern for all 8 constants
```

**Suggested change:**

```java
APICURIO_AVRO("apicurio_avro"),
CONFLUENT_AVRO("confluent_avro"),
APICURIO_JSONSCHEMA("apicurio_jsonschema"),
APICURIO_PROTOBUF("apicurio_protobuf"),
CSV("csv"),
XML("xml"),
JSON("json"),
BINARY("binary"),
```

**Reason:** The `@JsonProperty` annotations on the constants do nothing. Jackson uses `@JsonCreator` to read the value and `@JsonValue` to write it - neither one looks at the per-constant annotations. The wire name (e.g. `"apicurio_avro"`) only needs to appear once, in the constructor argument. Removing the duplicate also makes each constant one line instead of three.

KasparMetsa (changes requested):

pantelis17:
Functional Changes vs main
  
  1. notations.<name>.type is now a validated enum. Unknown values fail fast at startup instead of being silently skipped.
  2. errorHandling.*.handler now rejects unknown values with a clear error (was an NPE) and no longer accepts the short aliases stop/continue/retry.
  3. Unknown properties under a notation entry now fail at startup instead of being silently ignored.
  4. Spec/internal: `ksml-runner-spec.json` now models type/handler as explicit enums (output unchanged), plus a new ksml-docs-test module with schema-validation tests and added unit tests.
  
  Those changes where made in order to align the code logic with the ksml-runner-spec.json

ksml-runner/src/main/java/io/axual/ksml/runner/config/ErrorHandlingConfig.java
```diff
@@ -102,26 +108,29 @@ public static class ErrorTypeHandlingConfig {
...
             @JsonCreator
-            public static Handler forValues(String value) {
+            public static Handler forValue(String value) {
                 if (value == null) {
                     return null;
                 }
-
-                return switch (value) {
-                    case "continue", "continueOnFail" -> CONTINUE;
-                    case "stop", "stopOnFail" -> STOP;
```
KasparMetsa, line 121: Please confirm if this is what we want to do:

**What changed:** The old `Handler.forValues()` accepted both the short form and the canonical form for each value:

```java
// Old code
case "continue", "continueOnFail" -> CONTINUE;
case "stop", "stopOnFail" -> STOP;
case "retry", "retryOnFail" -> RETRY;
```

The new `Handler.forValue()` only accepts the canonical form. Any other value throws an error:

```java
// New code
for (final var handler : values()) {
    if (handler.jsonValue.equals(value)) {  // only "stopOnFail", "continueOnFail", "retryOnFail"
        return handler;
    }
}
throw new IllegalArgumentException("Unknown error handler: " + value + ...);
```
**Why this matters:** A user who has `handler: stop` (or `handler: continue`, `handler: retry`) in their `ksml-runner.yaml` will get a startup failure after upgrading. The runner will throw `Unknown error handler: stop`. This was working before the PR.

if this is what we want to do we need to update docs:

### Documentation table - needs one fix

`docs/reference/configuration-reference.md` line 155 shows the **default** value for `handler` as `stop` (the short form):

```markdown
| `handler` | String | stop | Error handling strategy (`continueOnFail`, `stopOnFail`, `retryOnFail`) |
```

After this PR, `stop` is no longer a valid value. The default should say `stopOnFail`.

**Suggested fix:**

```markdown
| `handler` | String | stopOnFail | Error handling strategy (`continueOnFail`, `stopOnFail`, `retryOnFail`) |
```

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #634, opened 2026-06-24
Ksml metrics help text

pantelis17:
# Add HELP text for KSML Prometheus metrics + fix ksml_app label mangling

## Summary

  System admins scraping KSML metrics saw uninformative auto-generated HELP text (e.g. # HELP ksml_execution_time_count Attribute exposed for management ksml:name=null,type=execution-time,attribute=Count) and a mangled ksml_app metric. This PR makes every ksml_ metric self-documenting and cleans up the app-info labels.

## Changes

  - HelpEnrichingCollector (new) — a MultiCollector that wraps the Prometheus JMX exporter and rewrites only the HELP text of ksml_ metrics, passing names, labels, values, and types through unchanged. Covers all four KSML metric families: ksml_app, ksml_execution_time, ksml_record_e2e_latency_{avg,min,max}_ms, and ksml_user_defined_{counter,meter,timer}, with correct units noted (durations in ms, rates per second).
  - PrometheusExport — wires the wrapper in via an extracted helper registerJmxCollectorWithHelpText(...): register the JmxCollector (so it initializes its operational metrics), unregister it, then register the wrapper around it.
  - default_config.yaml — fixed the app-info rule regex so ksml_app labels (name, version, build_time) are clean instead of carrying leftover JMX quotes and a >< artifact.
  - Docs — new reference/metrics-reference.md cataloguing all KSML metrics, their labels, units, and the Python user-metrics API; wired into mkdocs.yml and the references overview.

## Why this approach

  The obvious fix — adding help: to rules in default_config.yaml — does not work, verified against jmx_exporter 1.0.1:

  1. The exporter rejects any rule with help but no name (Must provide name, if help or labels are given), which fails config load and stops all metric export.
  2. A rule with an explicit name does not auto-generate labels from JMX bean properties, so KSML's dynamic labels (function_name, pipeline, namespace, custom_name, …) get dropped.

  Since rule-based HELP is incompatible with preserving those labels, rewriting HELP after collection is the only approach that documents metrics without losing labels. The register → unregister → register sequence preserves the exporter's own operational metrics (jmx_scrape_duration_seconds, jmx_config_reload_*). The app-info label fix lives in the config regex (not KsmlInfo) because the ObjectName.quote() quoting is required for a valid JMX ObjectName.

## Testing

  - HelpEnrichingCollectorTest (@Nested, 5 tests):
    - Through the real JMX exporter + shipped config: every KSML family gets enriched HELP while keeping its JMX-derived labels; and the PrometheusExport wiring keeps the exporter's operational metrics.
    - Rewrite logic with a stub delegate: counter type preserved (not downgraded to gauge), non-matching/already-described metrics passed through as the same instance, both predicate collect overloads and getPrometheusNames delegate correctly.
    - Coverage on HelpEnrichingCollector: 93% instructions, 88% branches, 100% methods.
  - AppInfoMetricTest: registers the real app-info MBean and scrapes through the real JmxCollector + shipped config, asserting ksml_app labels are clean (no quotes, no </>).

  Live verification (Docker: Kafka + schema registry + producer + example 17, scraping :9999/metrics):

  - Confirmed the bug existed on main (every ksml_ metric showed Attribute exposed for management …).
  - With the change: all ksml_ metrics show descriptive HELP, labels preserved, e.g.
  # HELP ksml_execution_time_count Execution time statistics of a KSML user function per invocation; durations are in milliseconds and rates are per second
  ksml_execution_time_count{function_name="pipelines_consume_avro_forEach",function_type="forEach",namespace="inspect",operation_name="...",pipeline="consume_avro"}
  18.0
  - ksml_app before vs after:
  before: build_time="\"2026-...Z\"><",name="\"KSML Runner\"",version="\"2.0.0-SNAPSHOT\""
  after:  build_time="2026-06-24T13:52:32Z",name="KSML Runner",version="2.0.0-SNAPSHOT"
  - Confirmed nothing else broke: all metric families still present (~same count), no metric left with the default/mangled help, and JVM/Kafka/operational metrics intact.

## Note

  The pre-existing ksml<type=user_defined_counter, ...> rule in default_config.yaml uses underscores but the real JMX type is user-defined-counter (hyphens), so user-defined counters currently export as gauges. Fixing it changes the metric type (_total suffix), so it's left out of this PR's scope — worth a separate ticket.

### DIFF
--- a/docs/reference/metrics-reference.md
+++ b/docs/reference/metrics-reference.md
@@ -1,0 +1,154 @@
+# Metrics Reference
+
+This document describes the metrics that the KSML Runner exposes for Prometheus, including the custom
+KSML metrics, their labels and units, and the other metrics that are available on the same endpoint.
+
+## Enabling the Metrics Endpoint
+
+Metrics are exposed over HTTP by the Prometheus JMX Exporter that is built into the KSML Runner. The
+endpoint is disabled by default and is enabled through the `prometheus` section of `ksml-runner.yaml`:
+
+```yaml
+ksml:
+  prometheus:
+    enabled: true
+    host: 0.0.0.0
+    port: 9999
+```
+
+See the [Configuration Reference](configuration-reference.md) for all available options. Once enabled,
+the metrics are served at:
+
+```
+http://<host>:<port>/metrics
+```
+
+Every metric carries a `# HELP` line describing what it measures, so the quickest way to discover the
+metrics exposed by a running pipeline is:
+
+```bash
+curl -s localhost:9999/metrics | grep '^# HELP'
+```
+
+## KSML Custom Metrics
+
+KSML publishes its own metrics under the `ksml_` prefix. Each metric is labelled so that values can be
+attributed to a specific function, pipeline or processor node.
+
+The metrics listed in this section are the complete set of metrics that KSML defines itself. Each one
+is given a descriptive `# HELP` line at runtime.
+
+### Function Execution Time
+
+`ksml_execution_time` measures the wall-clock time spent in a user function for every invocation. It is
+recorded for all function types (forEach, mapper, predicate, etc.).
+
+| Property | Value |
+|----------|-------|
+| Metric   | `ksml_execution_time` |
+| Type     | Timer |
+| Unit     | Durations in milliseconds, rates per second |
+| Labels   | `function_name`, `function_type`, `namespace`, `pipeline`, `operation_name` (and `step` or `branch` where applicable) |
+
+### Record End-to-End Latency
+
+The `ksml_record_e2e_latency_*_ms` gauges report the latency of a record measured from the source topic
+to a specific processor node, as reported by Kafka Streams and enriched with KSML context.
+
+| Property | Value |
+|----------|-------|
+| Metrics  | `ksml_record_e2e_latency_avg_ms`, `ksml_record_e2e_latency_min_ms`, `ksml_record_e2e_latency_max_ms` |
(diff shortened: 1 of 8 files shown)


docs/reference/metrics-reference.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Application Information
+
+`ksml_app` is a constant gauge (value `1`) that carries build and version information as labels.
+
+| Property | Value |
+|----------|-------|
+| Metric   | `ksml_app` |
+| Type     | Gauge |
+| Labels   | `app_id`, `name`, `version`, `build-time` |
```
KasparMetsa: ### Docs say label `build-time` but the actual label is `build_time`

**File:** `docs/reference/metrics-reference.md` line 131

The metrics reference lists the `ksml_app` labels as:

```markdown
| Labels | `app_id`, `name`, `version`, `build-time` |
```

But `default_config.yaml` defines the label with an underscore:

```yaml
labels:
  app_id: "$1"
  name: "$2"
  version: "$3"
  build_time: "$4"     # underscore, not hyphen
```

`AppInfoMetricTest` also asserts `labels.get("build_time")` (underscore). A user who queries Prometheus with `ksml_app{build-time="..."}` based on the docs will get no results.

**Suggested fix:**

```markdown
| Labels | `app_id`, `name`, `version`, `build_time` |
```

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #639, opened 2026-06-29
Added documentation about local sonar/jacoco analysis

pantelis17:


jeroenvandisseldorp (approved):

CONTRIBUTING.md
```diff
@@ -121,6 +121,65 @@ Those two numbers will match what SonarCloud reports.
 
 Per-module reports (lower numbers, own tests only) are still at `<module>/target/site/jacoco/index.html` if you need them.
 
+### Checking Diff Coverage (coverage of changed lines only)
+
+To check test coverage only for the lines changed in your branch compared to `main`, use [diff-cover](https://github.com/Bachmann1234/diff_cover). First install it:
+
+```shell
+pip install diff-cover
```
KasparMetsa: `pip install` will fail for anyone using Homebrew Python.  Homebrew only ships pip3.  pip3 is unambiguous - it always refers to Python 3. pip is a leftover convention from when Python 2 was common and pip meant Python 2's pip. On modern systems it either doesn't exist or points to the same thing as pip3. Always use pip3.

Also when using `pip3` use a venv, so we don't break any other dependencies on developer's system.

Alternatively, recommend to use pipx which installs CLI tools in isolated environments automatically and puts them in PATH without any venv activation step:
```
pipx install diff-cover
```
That is arguably the cleanest option for a tool like diff-cover - install once, use anywhere, no venv activation, no system pollution.

And add: "If you don't have pipx, install it with brew install pipx."

KasparMetsa (changes requested):

CONTRIBUTING.md
```diff
@@ -121,6 +121,65 @@ Those two numbers will match what SonarCloud reports.
...
+```
+
+> **Note:** On some systems `diff-cover` is installed to a user-local path (e.g. `~/Library/Python/3.9/bin/diff-cover` on macOS). If the command is not found after install, use the full path or add the directory to your `PATH`.
+
+Make sure you have run `mvn clean verify --no-transfer-progress` first to produce the per-module JaCoCo reports, then run:
+
+```shell
+diff-cover **/target/site/jacoco/jacoco.xml \
```
KasparMetsa:
```shell
diff-cover **/target/site/jacoco/jacoco.xml \
  --compare-branch origin/main \
  --src-roots **/src/main/java \
  --format html:diff-coverage.html
```

**Fix:**

```shell
diff-cover ksml-reporting/target/site/jacoco-aggregate/jacoco.xml \
  --compare-branch origin/main \
  --src-roots **/src/main/java \
  --format html:diff-coverage.html
```

**Reason:** The glob `**/target/site/jacoco/jacoco.xml` expands to the per-module XML
files - one per module, each containing only that module's own test results. The file
`ksml-reporting/target/site/jacoco-aggregate/jacoco.xml` is the single aggregated XML
that merges coverage data from all modules together. This is the exact file SonarCloud
is configured to read (set in `pom.xml` via `sonar.coverage.jacoco.xmlReportPaths`).

Both commands produce the same diff-coverage percentage in practice, because diff-cover
only looks at lines that changed in the branch - and a changed file belongs to exactly
one module, so its coverage data appears in both the per-module XML and the aggregate
XML. However, pointing at the aggregate XML makes the intent explicit: you are running
against the same data source as SonarCloud, so the number you see is guaranteed to match
what SonarCloud will report.

Verified: running both commands on PR #640 produced identical output.

KasparMetsa (changes requested):

CONTRIBUTING.md
```diff
@@ -4,17 +4,24 @@ Thank you for considering contributing to this project! Before making any change
...
 Please also make sure to read and follow our [Code of Conduct](CODE_OF_CONDUCT.md) in all interactions with the project.
 
-## Merge Request Process
+## Pull Request Process
 
 1. Fork the repository and create a feature or bugfix branch.
 2. Ensure your code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).
 2. Document your code clearly and thoroughly.
-3. Maintain at least 70% test coverage. Include a unit or integration test that reproduces any bug you're fixing.
+3. Include a unit or integration test that reproduces any bug you're fixing.
```
KasparMetsa:
```suggestion
3. Maintain at least 60% test coverage. Include a unit or integration test that reproduces any bug you're fixing.
```

CONTRIBUTING.md
```diff
@@ -4,17 +4,24 @@ Thank you for considering contributing to this project! Before making any change
...
 6. Update the [`README.md`](README.md) when you introduce interface changes (e.g., new environment variables, file paths, etc.).
 7. Submit a pull request from your fork to the main repository.
 8. A pull request may be merged after approval from two developers. If you do not have merge permissions, ask the second reviewer to merge it for you.
 
+## Before Sending for Review
+
+The following two steps are **mandatory** before asking someone to review:
+
+1. Run the SonarCloud analysis in your IDE and resolve any reported issues. See [Sonar IDE integration](#sonar-ide-integration) below.
+2. Run `diff-cover` as described above and confirm that coverage of your changes is above 60%. See [Checking Diff Coverage](#checking-diff-coverage-coverage-of-changed-lines-only) below.
```
KasparMetsa:
```suggestion
2. Run `diff-cover` and confirm that coverage of your changes is above 60%.
   See [Checking Diff Coverage](#checking-diff-coverage-coverage-of-changed-lines-only) below.
```

CONTRIBUTING.md
```diff
@@ -121,6 +128,62 @@ Those two numbers will match what SonarCloud reports.
...
+diff-cover ksml-reporting/target/site/jacoco-aggregate/jacoco.xml \
+  --compare-branch origin/main \
+  --src-roots **/src/main/java \
+  --format html:diff-coverage.html
+```
+
+Open the report:
+
+```shell
+open diff-coverage.html
```
KasparMetsa:
```suggestion
open diff-coverage.html        # macOS
xdg-open diff-coverage.html   # Linux
```

KasparMetsa (changes requested):

CONTRIBUTING.md
pantelis17: open works fine for linux too, why to use this? Also, our previous doc was only for macOS e.g. in installation of pipx

KasparMetsa (approved):

PR #644, opened 2026-06-30
Add ksml query rest tests

pantelis17:
# Add unit tests for the ksml-query REST module

## What this PR does

This PR adds unit tests to the `ksml-query` module. The module runs a REST
server that lets you query Kafka Streams state stores (both local and remote).
Before this change it had almost no test coverage.

While writing the tests, a few real bugs and some dead code showed up in the
source. Those are fixed here too.

### New tests

Added test classes for the REST resources, the client, the server and the data
beans:

- `KeyValueStoreResourceTest`
- `WindowedKeyValueStoreResourceTest`
- `RestClientTest`
- `RestServerTest`
- `RestServerExceptionMapperTest`
- `LivenessResourceTest`
- `ReadyResourceTest`
- `StartupResourceTest`
- `GlobalStateTest`
- `HostDiscoveryTest`
- `KeyValueBeansTest`
- `WindowedKeyValueBeansTest`

The tests use JUnit 5, Mockito and AssertJ.

### Bug fixes found while testing

**1. Wrong host in the remote query URL**

The URL used the whole host object instead of only the host name. This produced
a broken address. It is now fixed in both key-value and windowed resources.

```java
// before
"http://" + metadataForKey.activeHost() + ":" + ...
// after
"http://" + metadataForKey.activeHost().host() + ":" + ...
```

**2. Wrong path for the windowed store**

The windowed resource was calling the `keyvalue` path and did not send the
timestamp. It now calls the correct `windowed` path and adds the timestamp.

```java
// before
".../state/keyvalue/" + storeName + "/local/get/" + key
// after
".../state/windowed/" + storeName + "/local/get/" + key + "/" + timestamp
```

**3. Endpoint casing**

The key-value path is now `keyValue` so it matches the resource mapping.

### Cleanup

- Removed unused REST client helpers and a duplicate `getHostIPForDiscovery`
  method from `Utils` and `RestClient`.
- Fixed 4 SonarCloud issues.

### Build

- Added the JUnit 5, Mockito and Mockito-JUnit test dependencies to
  `ksml-query/pom.xml`.

## Why

Without tests these bugs were easy to miss. The wrong host and wrong path meant
remote store queries would fail in a multi-instance setup. Adding tests gives us
a safety net and documents how the REST layer is expected to behave.


ksml-query/src/main/java/io/axual/ksml/rest/server/RestClient.java
KasparMetsa: I know you didn't change this but it is related to other changes.


## The problem

All three methods make an HTTP call and wait up to 4 seconds for a response. If the response takes too long (timeout) or fails, they need to return something.

**The first two methods return an empty list on failure:**

```java
// getRemoteKeyValueBeans - returns empty list on timeout
return new KeyValueBeans();   // an empty list, safe to use

// getRemoteWindowKeyValueBeans - returns empty list on timeout
return new WindowedKeyValueBeans();   // an empty list, safe to use
```

**The third method returns null on failure:**

```java
// getRemoteKeyValueBean - returns null on timeout
return null;   // <-- this is the problem
```

The caller in `KeyValueStoreResource` then does this:

```java
var result = restClient.getRemoteKeyValueBean(url, KeyValueBean.class);
log.info("Store data from remote store at {} == {}", url, result);
return result;   // returns null to the REST framework
```

It gets `null` back and returns it directly to the REST framework (JAX-RS). It never checks whether `result` is null.

---

## What the client sees

When a REST method returns `null` in JAX-RS, the framework sends back an HTTP `204 No Content` response - a successful response with an empty body.

So the client's experience is:

1. Client asks: "What is the value for key X?"
2. The remote instance is slow or down
3. The timeout fires, `null` is returned
4. Client receives: **HTTP 204 - success, no data**

The client has no way to know that something went wrong. They just see an empty success response, which looks the same as "the key does not exist". This is a silent failure.

---

## What it should do instead

When a single-key lookup fails (timeout or error), the correct response is to tell the client that the service is unavailable. In JAX-RS, this is done by throwing a `ServiceUnavailableException`, which produces an HTTP `503 Service Unavailable` response. The client can then retry or show an error.

The other two methods (that fetch lists) return empty lists on failure because merging partial results is fine for a full-scan query. But for a single-key lookup, returning null is wrong because null and "key not found" look the same to the caller.

---

## The fix

**Current code** (`RestClient.java`):

```java
public <T extends KeyValueBean> T getRemoteKeyValueBean(String url, Class<T> resultClass, Duration duration) {
    try {
        Future<T> storeDataFuture = getRESTClient().target(url)
                .request(MediaType.APPLICATION_JSON)
                .async()
                .get(resultClass);
        return storeDataFuture.get(duration.toMillis(), TimeUnit.MILLISECONDS);
    } catch (InterruptedException | ExecutionException _) {
        log.warn(INTERRUPTED_MESSAGE, url);
        Thread.currentThread().interrupt();
    } catch (TimeoutException _) {
        log.warn(TIMEOUT_MESSAGE, url);
    }
    return null;   // <-- problem: null causes a silent 204 response
}
```

**Suggested fix:**

```java
public <T extends KeyValueBean> T getRemoteKeyValueBean(String url, Class<T> resultClass, Duration duration) {
    try {
        Future<T> storeDataFuture = getRESTClient().target(url)
                .request(MediaType.APPLICATION_JSON)
                .async()
                .get(resultClass);
        return storeDataFuture.get(duration.toMillis(), TimeUnit.MILLISECONDS);
    } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
        throw new ServiceUnavailableException("Remote store query was interrupted");
    } catch (ExecutionException e) {
        throw new ServiceUnavailableException("Remote store query failed: " + e.getMessage());
    } catch (TimeoutException e) {
        throw new ServiceUnavailableException("Remote store query timed out");
    }
}
```

With this fix, when the remote call fails, JAX-RS receives a `ServiceUnavailableException` and automatically sends an HTTP `503 Service Unavailable` response. The client knows something went wrong and can decide what to do.

The test for this method should also be updated to match:

**Current test** (`RestClientTest.java`):

```java
@Test
@DisplayName("getRemoteKeyValueBean returns null when the call times out")
void singleBeanTimeout() throws Exception {
    stubChainReturning(mockFutureThrowing(new TimeoutException()));

    assertThat(restClient.getRemoteKeyValueBean(URL, KeyValueBean.class)).isNull();
}
```

**Suggested fix:**

```java
@Test
@DisplayName("getRemoteKeyValueBean throws ServiceUnavailableException when the call times out")
void singleBeanTimeout() throws Exception {
    stubChainReturning(mockFutureThrowing(new TimeoutException()));

    assertThatThrownBy(() -> restClient.getRemoteKeyValueBean(URL, KeyValueBean.class))
            .isInstanceOf(ServiceUnavailableException.class);
}
```

ksml-query/src/test/java/io/axual/ksml/rest/server/RestClientTest.java
```diff
@@ -0,0 +1,178 @@
...
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class RestClientTest {
+
+    private static final String URL = "http://remote:9090/state/keyvalue/store/local/all";
```
KasparMetsa: ### `RestClientTest.URL` uses the old buggy path casing

One of the three bugs this PR fixes is the wrong casing: `keyvalue` → `keyValue`. But `RestClientTest` still uses a constant with the old wrong casing:

**PR code:**

```java
// RestClientTest.java
private static final String URL = "http://remote:9090/state/keyvalue/store/local/all";
//                                                           ^^^^^^^^^
//                                                           wrong - this is the casing that was just fixed
```

**Suggested fix:**

```java
private static final String URL = "http://remote:9090/state/keyValue/store/local/all";
```

**Reason:** `RestClientTest` tests HTTP client behavior, not URL construction, so the wrong casing doesn't cause a test failure. But the constant is copy-paste bait - a developer reading this file would not know the correct URL casing. The correct form is `keyValue`, as asserted by the routing tests in `KeyValueStoreResourceTest`:

```java
assertThat(url.getValue()).isEqualTo("http://other:9090/state/keyValue/" + STORE + "/local/get/k1");
```

ksml-query/src/main/java/io/axual/ksml/rest/server/RestClient.java
KasparMetsa: ### `InterruptedException` path in `RestClient` is not covered by tests

`RestClient` catches `InterruptedException` and restores the interrupt flag:

```java
// RestClient.java (existing in the async methods)
} catch (InterruptedException | ExecutionException _) {
    log.warn("Store data fetch from {} was interrupted", url);
    Thread.currentThread().interrupt();
}
```

`RestClientTest` covers `TimeoutException` and `ExecutionException` paths but has no test for `InterruptedException`. Since `InterruptedException` also calls `Thread.currentThread().interrupt()`, it is worth verifying that the interrupt flag is restored.

**Suggested addition to `RestClientTest`:**

```java
@Test
@DisplayName("getRemoteKeyValueBeans restores the interrupt flag when interrupted")
void keyValueBeansInterrupted() throws Exception {
    stubChainReturning(mockFutureThrowing(new InterruptedException()));

    restClient.getRemoteKeyValueBeans(URL);

    assertThat(Thread.currentThread().isInterrupted()).isTrue();
    Thread.interrupted(); // clear for other tests
}
```

**Reason:** Interrupt handling is easy to break silently - forgetting `Thread.currentThread().interrupt()` would swallow the interrupt without any test failure. One test covers the behavior and catches future regressions.

KasparMetsa (changes requested):

ksml-query/src/main/java/io/axual/ksml/rest/server/RestClient.java
```diff
@@ -132,13 +133,17 @@ public <T extends KeyValueBean> T getRemoteKeyValueBean(String url, Class<T> res
                     .async() //returns asap
                     .get(resultClass);
             return storeDataFuture.get(duration.toMillis(), TimeUnit.MILLISECONDS); //blocks until timeout
-        } catch (InterruptedException | ExecutionException _) {
+        } catch (InterruptedException _) {
             log.warn(INTERRUPTED_MESSAGE, url);
             Thread.currentThread().interrupt();
+            throw new ServiceUnavailableException("Remote store query was interrupted");
+        } catch (ExecutionException _) {
+            log.warn(INTERRUPTED_MESSAGE, url);
```
KasparMetsa: ### 1. `ExecutionException` catch block logs the wrong message in `RestClient.getRemoteKeyValueBean`

**File:** `ksml-query/src/main/java/io/axual/ksml/rest/server/RestClient.java`

**PR code:**

```java
} catch (InterruptedException _) {
    log.warn(INTERRUPTED_MESSAGE, url);   // "Store data fetch from {} was interrupted"
    Thread.currentThread().interrupt();
    throw new ServiceUnavailableException("Remote store query was interrupted");
} catch (ExecutionException _) {
    log.warn(INTERRUPTED_MESSAGE, url);   // <-- wrong: also says "was interrupted"
    throw new ServiceUnavailableException("Remote store query failed");
} catch (TimeoutException _) {
    log.warn(TIMEOUT_MESSAGE, url);       // "Store data fetch from {} timed out"
    throw new ServiceUnavailableException("Remote store query timed out");
}
```

**Problem:** When an `ExecutionException` occurs (the remote call failed for any reason other than a timeout), the log says `"Store data fetch from <url> was interrupted"`. This is the wrong message - `INTERRUPTED_MESSAGE` was copy-pasted from the `InterruptedException` block when the combined catch was split. An operator reading the log will think the thread was interrupted, when in fact the call failed with an execution error.

**Suggested fix:**

```java
private static final String INTERRUPTED_MESSAGE = "Store data fetch from {} was interrupted";
private static final String FAILED_MESSAGE = "Store data fetch from {} failed";   // add this
private static final String TIMEOUT_MESSAGE = "Store data fetch from {} timed out";

// ...

} catch (InterruptedException _) {
    log.warn(INTERRUPTED_MESSAGE, url);
    Thread.currentThread().interrupt();
    throw new ServiceUnavailableException("Remote store query was interrupted");
} catch (ExecutionException _) {
    log.warn(FAILED_MESSAGE, url);   // correct message
    throw new ServiceUnavailableException("Remote store query failed");
} catch (TimeoutException _) {
    log.warn(TIMEOUT_MESSAGE, url);
    throw new ServiceUnavailableException("Remote store query timed out");
}
```

**Reason:** The three error types have three distinct causes. A log that says "was interrupted" when the cause was actually an execution failure makes debugging harder - an operator looking at the log will investigate thread interruption when they should be looking at the remote call failure instead.

KasparMetsa (changes requested):

ksml-query/src/main/java/io/axual/ksml/rest/server/RestClient.java
KasparMetsa, line 79: ## The same bug that was just fixed still lives in the two list methods

The PR fixed the interrupt handling and log message in `getRemoteKeyValueBean` (the single-key method). But the two list methods, `getRemoteKeyValueBeans` and `getRemoteWindowKeyValueBeans`, still use the old combined catch. The new `FAILED_MESSAGE` constant that was added for the fix is only used by the single-key method.

**PR code** (both list methods have this exact pattern):

```java
public KeyValueBeans getRemoteKeyValueBeans(String url, Duration duration) {
    try {
        Future<KeyValueBeans> storeDataFuture = getRESTClient().target(url)
                .request(MediaType.APPLICATION_JSON)
                .async() //returns asap
                .get(KeyValueBeans.class);
        return storeDataFuture.get(duration.toMillis(), TimeUnit.MILLISECONDS); //blocks until timeout
    } catch (InterruptedException | ExecutionException _) {
        log.warn(INTERRUPTED_MESSAGE, url);   // <-- wrong message when it was an ExecutionException
        Thread.currentThread().interrupt();   // <-- sets the interrupt flag even when NOT interrupted
    } catch (TimeoutException _) {
        log.warn(TIMEOUT_MESSAGE, url);
    }

    return new KeyValueBeans();
}
```

**Problem:** There are two issues, and the second one is a real bug, not just a cosmetic one.

1. Wrong log message. When the cause is an `ExecutionException` (the remote call failed for a reason other than a timeout), the log still says "Store data fetch from {} was interrupted". This is the exact problem that was already fixed in `getRemoteKeyValueBean`. An operator reading the log will look for a thread interruption when the real cause is a failed remote call.

2. Wrong interrupt flag. On an `ExecutionException` the code calls `Thread.currentThread().interrupt()`. This sets the interrupt flag on a thread that was never interrupted. These methods run on a JAX-RS server worker thread from a pool. The flag stays set after the request finishes. The next blocking call on that same pooled thread can then throw `InterruptedException` for no real reason. This is a hidden, hard-to-reproduce failure.

**Suggested fix** (apply to both `getRemoteKeyValueBeans` and `getRemoteWindowKeyValueBeans`):

```java
public KeyValueBeans getRemoteKeyValueBeans(String url, Duration duration) {
    try {
        Future<KeyValueBeans> storeDataFuture = getRESTClient().target(url)
                .request(MediaType.APPLICATION_JSON)
                .async() //returns asap
                .get(KeyValueBeans.class);
        return storeDataFuture.get(duration.toMillis(), TimeUnit.MILLISECONDS); //blocks until timeout
    } catch (InterruptedException _) {
        log.warn(INTERRUPTED_MESSAGE, url);
        Thread.currentThread().interrupt();   // correct: only when actually interrupted
    } catch (ExecutionException _) {
        log.warn(FAILED_MESSAGE, url);         // correct message, no interrupt flag
    } catch (TimeoutException _) {
        log.warn(TIMEOUT_MESSAGE, url);
    }

    return new KeyValueBeans();
}
```

**Reason:** The three exception types have three different causes and must be handled separately. `Thread.currentThread().interrupt()` belongs only in the `InterruptedException` branch. Keeping the old combined catch means the code the reviewer already asked to fix is still present in two of the three methods, and it can corrupt the interrupt state of a shared request thread. It is not an option to fix one method and leave its two siblings broken.

ksml-query/src/test/java/io/axual/ksml/rest/server/RestClientTest.java
```diff
@@ -0,0 +1,224 @@
...
+    @Test
+    @DisplayName("getRemoteKeyValueBeans returns empty beans when the call times out")
+    void keyValueBeansTimeout() throws Exception {
+        stubChainReturning(mockFutureThrowing(new TimeoutException()));
+
+        assertThat(restClient.getRemoteKeyValueBeans(URL).elements()).isEmpty();
+    }
+
+    @Test
+    @DisplayName("getRemoteKeyValueBeans returns empty beans when the call fails")
```
KasparMetsa: previous bug fix must be done before this = https://github.com/Axual/ksml/pull/644/changes#r3569317301

The new `keyValueBeansInterrupted` test proves the interrupt flag is set after an `InterruptedException`. But there is no matching test that proves the flag is left clear after an `ExecutionException`. This is exactly why the previously written bug can stay hidden.

**PR code:**

```java
@Test
@DisplayName("getRemoteKeyValueBeans returns empty beans when the call fails")
void keyValueBeansExecutionError() throws Exception {
    stubChainReturning(mockFutureThrowing(new ExecutionException(new RuntimeException("boom"))));

    assertThat(restClient.getRemoteKeyValueBeans(URL).elements()).isEmpty();
}
```

**Problem:** The test only checks the returned list is empty. With the current code the interrupt flag is wrongly set here, and the test does not notice. Worse, this test does not clear the flag afterwards (the interrupt tests do call `Thread.interrupted()` to clean up). So the flag can leak into whatever test runs next on the same thread.

**Suggested fix** (after the previous comment fix, add the assertion so the behavior stays correct):

```java
@Test
@DisplayName("getRemoteKeyValueBeans returns empty beans and does not set the interrupt flag when the call fails")
void keyValueBeansExecutionError() throws Exception {
    stubChainReturning(mockFutureThrowing(new ExecutionException(new RuntimeException("boom"))));

    assertThat(restClient.getRemoteKeyValueBeans(URL).elements()).isEmpty();
    assertThat(Thread.currentThread().isInterrupted()).isFalse();
}
```

**Reason:** A failed remote call is not an interruption, so the flag must stay clear. Adding one assertion locks in the previous bug fix and stops future regressions. It also keeps the test thread clean for the tests that run after it.

ksml-query/src/test/java/io/axual/ksml/rest/server/RestClientTest.java
```diff
@@ -0,0 +1,224 @@
...
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class RestClientTest {
+
+    private static final String URL = "http://remote:9090/state/keyValue/store/local/all";
```
KasparMetsa:
```java
private static final String URL = "http://remote:9090/state/keyValue/store/local/all";
```

**Problem:** This URL ends in `/local/all`, which is the list endpoint. The same constant is reused by the single-key tests (`singleBeanSuccess`, `singleBeanTimeout`, `singleBeanExecutionError`, `singleBeanInterrupted`), which in real code call `/local/get/{key}`. The mock does not care about the URL text, so nothing fails, but a reader cannot tell the correct single-key URL shape from this file.

**Suggested fix:**

```java
private static final String LIST_URL = "http://remote:9090/state/keyValue/store/local/all";
private static final String SINGLE_URL = "http://remote:9090/state/keyValue/store/local/get/k1";
```

Then use `SINGLE_URL` in the `singleBean*` tests and `LIST_URL` in the list tests.

**Reason:** Tests double as documentation for how the REST layer is called. A single-key test that points at the `/all` path is misleading. This is only a clarity issue, so it is low priority.

ksml-query/src/main/java/io/axual/ksml/rest/server/RestClient.java
KasparMetsa:
```java
    /**
     * Fetch remote KeyValueBeans using REST call
     *
     * @param url remote keyvalue endpoint URL
     * @return the StoreData
     */
    public WindowedKeyValueBeans getRemoteWindowedKeyValueBeans(String url) { ... }

    /**
     * Fetch remote KeyValueBeans using REST call
     *
     * @param url remote keyvalue endpoint URL
     * @return the StoreData
     */
    public <T extends KeyValueBean> T getRemoteKeyValueBean(String url, Class<T> resultClass) { ... }
    ```

    **Problem:** The windowed method returns `WindowedKeyValueBeans`, not `KeyValueBeans`, so its text is wrong. The single-key method returns one bean, has an extra `resultClass` parameter that is not documented, and now throws `ServiceUnavailableException` on failure, which the text does not mention. `@return the StoreData` names a type that does not exist here.

    **Suggested fix:**

    ```java
    /**
     * Fetch a page of windowed key-value entries from a remote instance over REST.
     *
     * @param url remote windowed-store endpoint URL
     * @return the windowed entries, or an empty set if the call fails or times out
     */
    public WindowedKeyValueBeans getRemoteWindowedKeyValueBeans(String url) { ... }

    /**
     * Fetch a single key-value entry from a remote instance over REST.
     *
     * @param url         remote single-key endpoint URL
     * @param resultClass the concrete bean type to deserialize into
     * @return the fetched bean
     * @throws jakarta.ws.rs.ServiceUnavailableException if the remote call fails, times out or is interrupted
     */
    public <T extends KeyValueBean> T getRemoteKeyValueBean(String url, Class<T> resultClass) { ... }
    ```

    **Reason:** Javadoc that names the wrong return type, skips a parameter, and hides the fact that the method now throws is worse than no Javadoc, because a caller trusts it. Match the text to each method, or drop the block if it adds nothing.

File: ksml-query/src/test/java/io/axual/ksml/rest/server/RestServerTest.java [hunk 5]
```diff
@@ -0,0 +1,86 @@
...
+            server.initGlobalQuerier(querier);
+
+            assertThat(GlobalState.INSTANCE.querier()).isSameAs(querier);
+            assertThat(GlobalState.INSTANCE.hostInfo()).isSameAs(hostInfo);
+        }
+    }
+
+    @Test
+    @DisplayName("The server binds locally, serves its registered resources and shuts down cleanly")
+    void servesRegisteredResources() throws Exception {
```
KasparMetsa:
```java
    private static int freePort() throws Exception {
        try (var socket = new ServerSocket(0)) {
            return socket.getLocalPort();
        }
    }

    @Test
    void servesRegisteredResources() throws Exception {
        final var hostInfo = new HostInfo("localhost", freePort());   // port is now free, but not held

        try (var server = new RestServer(hostInfo);
             var client = ClientBuilder.newClient()) {
            server.start();
            // ...
        }
    }
    ```

    **Problem:** `freePort()` opens a socket, reads its port, then closes the socket right away. Between that close and `server.start()` binding the same port, another process can take it. Then `start()` fails and the test flakes. The gap is small, but CI runs many things at once, so it can happen.

    **Suggested fix:** let the server itself pick the port. Pass port `0` so the OS assigns a free one at bind time, and read the real port back from the server. This needs a small accessor on `RestServer` that returns the port the Grizzly listener actually bound to:

    ```java
    // In RestServer: expose the port the server actually bound to.
    public int boundPort() {
        return server.getListener("grizzly").getPort();
    }

    // In the test: no freePort() helper, no gap.
    @Test
    void servesRegisteredResources() throws Exception {
        try (var server = new RestServer(new HostInfo("localhost", 0));
             var client = ClientBuilder.newClient()) {
            server.start();

            final var status = client.target("http://localhost:" + server.boundPort() + "/startup")
                    .request()
                    .get(Response.class)
                    .getStatus();

            assertThat(status).isEqualTo(500);
        }
    }
    ```

    **Reason:** With port `0` there is no window where the port is free but unheld, because the OS assigns it at the moment the server binds. This removes the race for good. It costs one small getter on `RestServer`. Because the current race is rare and low risk, this is a note, not a blocker.

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #646, opened 2026-07-01
Add unit test to ksml-data module

pantelis17:
# Add unit tests to `ksml-data` module

## What this PR does

Adds unit tests for the `ksml-data` module, which had no tests before. It also includes some runner code refactoring that was done to make the runner logic easier to test in a future PR.

---

### New tests in `ksml-data` (13 new files)

| Area | New test classes |
|---|---|
| Comparison | `EqualityFlagsTest`, `EqualityTest` |
| Exception | `DataExceptionTest` |
| Notation | `NotationTest`, `NotationStub` |
| Schema types | `DataSchemaConstantsTest`, `UnionSchemaMemberTest` |
| Data types | `UnionTypeMemberTest`, `UnresolvedTypeTest` |
| Utilities | `AssignableUtilTest`, `EqualUtilTest`, `JsonNodeUtilTest` |
| Value | `TupleTest` |

### Expanded tests in `ksml-data` (12 existing files extended)

`NotationProviderTest`, `EnumSchemaTest`, `FixedSchemaTest`, `ListSchemaTest`, `MapSchemaTest`, `StructFieldTest`, `StructSchemaTest`, `UnionSchemaTest`, `EnumTypeTest`, `StructTypeTest`, `UnionTypeTest`, `ConvertUtilTest`

### DIFF
--- a/ksml-data/src/test/java/io/axual/ksml/data/notation/NotationProviderTest.java
+++ b/ksml-data/src/test/java/io/axual/ksml/data/notation/NotationProviderTest.java
@@ -44,4 +44,49 @@ class NotationProviderTest {
         assertThat(provider.vendorName()).isNull();
         assertThat(provider.notationName()).isEqualTo("avro");
     }
+
+    @Test
+    @DisplayName("name() combines vendor and notation, uses notation alone when no vendor, and is null without a notation name")
+    void nameCombining() {
+        final var vendored = new NotationProvider() {
+            @Override
+            public String notationName() {
+                return "avro";
+            }
+
+            @Override
+            public String vendorName() {
+                return "confluent";
+            }
+
+            @Override
+            public Notation createNotation(NotationContext notationContext) {
+                return null;
+            }
+        };
+        assertThat(vendored.name()).isEqualTo("confluent_avro");
+
+        final var plain = new NotationProvider() {
+            @Override
+            public String notationName() {
+                return "json";
+            }
+
+            @Override
+            public Notation createNotation(NotationContext notationContext) {
+                return null;
+            }
+        };
+        assertThat(plain.name()).isEqualTo("json");
+
+        final var anonymous = new NotationProvider() {
+            @Override
+            public Notation createNotation(NotationContext notationContext) {
+                return null;
+            }
+        };
+        assertThat(anonymous.name()).isNull();
+        // The no-arg createNotation() default delegates to createNotation(null)
+        assertThat(anonymous.createNotation()).isNull();
+    }
 }
--- a/ksml-data/src/test/java/io/axual/ksml/data/notation/NotationProviderTest.java
+++ b/ksml-data/src/test/java/io/axual/ksml/data/notation/NotationProviderTest.java
@@ -56,7 +56,7 @@ class NotationProviderTest {
 
             @Override
             public String vendorName() {
-                return "confluent";
+                return "axual";
             }
 
             @Override
@@ -64,7 +64,7 @@ class NotationProviderTest {
                 return null;
             }
         };
-        assertThat(vendored.name()).isEqualTo("confluent_avro");
+        assertThat(vendored.name()).isEqualTo("axual_avro");
 
         final var plain = new NotationProvider() {
             @Override
@@ -73,10 +73,16 @@ class NotationProviderTest {
             }
 
             @Override
+            public String vendorName() {
+                return null; // explicit: no vendor
+            }
+
+            @Override
             public Notation createNotation(NotationContext notationContext) {
                 return null;
             }
         };
+        assertThat(plain.vendorName()).isNull(); // documents the assumption name() relies on
         assertThat(plain.name()).isEqualTo("json");
 
         final var anonymous = new NotationProvider() {
--- a/ksml-data/src/test/java/io/axual/ksml/data/schema/DataSchemaConstantsTest.java
+++ b/ksml-data/src/test/java/io/axual/ksml/data/schema/DataSchemaConstantsTest.java
@@ -23,8 +23,6 @@ package io.axual.ksml.data.schema;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
-import java.lang.reflect.Constructor;
-
 import static org.assertj.core.api.Assertions.assertThat;
 
 class DataSchemaConstantsTest {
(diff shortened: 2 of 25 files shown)


ksml-data/src/test/java/io/axual/ksml/data/notation/NotationProviderTest.java
```diff
@@ -44,4 +44,49 @@ public Notation createNotation(NotationContext notationContext) {
...
+    void nameCombining() {
+        final var vendored = new NotationProvider() {
+            @Override
+            public String notationName() {
+                return "avro";
+            }
+
+            @Override
+            public String vendorName() {
+                return "confluent";
```
pantelis17: Remove this reference

ksml-data/src/test/java/io/axual/ksml/data/schema/DataSchemaConstantsTest.java
```diff
@@ -0,0 +1,48 @@
...
+    void isType() {
+        assertThat(DataSchemaConstants.isType(DataSchemaConstants.STRING_TYPE)).isTrue();
+        assertThat(DataSchemaConstants.isType(DataSchemaConstants.INTEGER_TYPE)).isTrue();
+        assertThat(DataSchemaConstants.isType(DataSchemaConstants.UNION_TYPE)).isTrue();
+        assertThat(DataSchemaConstants.isType(DataSchemaConstants.STRUCT_TYPE)).isTrue();
+        assertThat(DataSchemaConstants.isType("not-a-type")).isFalse();
+    }
+
+    @Test
+    @DisplayName("Constants carry their documented values")
```
KasparMetsa: `DataSchemaConstantsTest.constants()` tests literal constant values, not behavior

**PR code:**

```java
@Test
@DisplayName("Constants carry their documented values")
void constants() {
    final var noTag = DataSchemaConstants.NO_TAG;
    final var namespace = DataSchemaConstants.DATA_SCHEMA_KSML_NAMESPACE;
    assertThat(noTag).isEqualTo(-1);
    assertThat(namespace).isEqualTo("io.axual.ksml.data");
}
```

**Problem:** These assertions verify that `NO_TAG == -1` and the namespace string equals a specific value. Both values are visible directly in the source code. The test adds no information beyond what `DataSchemaConstants.java` already says. If a future developer changes `NO_TAG` intentionally (say, to `Integer.MIN_VALUE`), the test fails - but it has no way to tell them whether the change is allowed or not.

**Suggested fix:** Remove the `constants()` test entirely. The `isType()` test in the same class already exercises the constants meaningfully (by passing `DataSchemaConstants.STRING_TYPE` into `isType()`). If the concern is that constants must not change accidentally, a comment in the production code is more appropriate than a test asserting a magic number.

If keeping the test, at minimum document why the exact values matter:

```java
@Test
@DisplayName("NO_TAG sentinel is negative so it cannot collide with any Avro schema tag (which are non-negative)")
void noTagSentinelIsNegative() {
    assertThat(DataSchemaConstants.NO_TAG).isNegative();
}
```

**Reason:** A test that only passes when the source code says what it says is testing the compiler, not the logic. It fails on every intentional change and gives no useful failure message.

ksml-data/src/test/java/io/axual/ksml/data/notation/NotationProviderTest.java
```diff
@@ -44,4 +44,49 @@ public Notation createNotation(NotationContext notationContext) {
...
+            }
+
+            @Override
+            public Notation createNotation(NotationContext notationContext) {
+                return null;
+            }
+        };
+        assertThat(vendored.name()).isEqualTo("confluent_avro");
+
+        final var plain = new NotationProvider() {
```
KasparMetsa, line 69: **PR code:**

```java
final var plain = new NotationProvider() {
    @Override
    public String notationName() {
        return "json";
    }

    @Override
    public Notation createNotation(NotationContext notationContext) {
        return null;
    }
};
assertThat(plain.name()).isEqualTo("json");
```

**Problem:** The `plain` provider relies on the default `vendorName()` returning `null` to produce `"json"` from `name()`. But `vendorName()` is not overridden, and there is no assertion that it returns `null`. If the default `vendorName()` implementation ever changes (e.g., to return an empty string instead of null), `name()` might produce `"_json"` or similar, the assertion `isEqualTo("json")` fails, but the test gives no clue about which method is wrong.

**Suggested fix:**

```java
final var plain = new NotationProvider() {
    @Override
    public String notationName() { return "json"; }

    @Override
    public String vendorName() { return null; }  // explicit: no vendor

    @Override
    public Notation createNotation(NotationContext notationContext) { return null; }
};
assertThat(plain.vendorName()).isNull();  // documents the assumption
assertThat(plain.name()).isEqualTo("json");
```

**Reason:** The test is implicitly depending on a default behavior it never checks. Making the assumption explicit (`vendorName()` returns null) turns a fragile implicit dependency into a documented contract.

ksml-data/src/test/java/io/axual/ksml/data/util/ConvertUtilTest.java
```diff
@@ -249,4 +269,252 @@ void convertStringToDataObject_invalidBoolean_honoursAllowFail() {
...
+    @DisplayName("convert: a value is converted to the first compatible member of a union type")
+    void convertToUnionMember() {
+        final var union = new UnionType(new UnionType.Member(DataLong.DATATYPE), new UnionType.Member(DataString.DATATYPE));
+
+        final var result = converter.convert(union, new DataInteger(5));
+
+        assertThat(result).isEqualTo(new DataLong(5L));
+    }
+
+    private static Notation notationConvertingTo(DataObject fixed) {
```
KasparMetsa, line 410: `ConvertUtilTest.notationConvertingTo` - 40-line anonymous stub could be a named inner class

**PR code:**

```java
private static Notation notationConvertingTo(DataObject fixed) {
    return new Notation() {
        @Override
        public Notation.SchemaUsage schemaUsage() { return null; }

        @Override
        public DataType defaultType() { return null; }

        @Override
        public String name() { return "stub"; }

        @Override
        public String filenameExtension() { return null; }

        @Override
        public Serde<Object> serde(DataType type, boolean isKey) { return null; }

        @Override
        public Notation.Converter converter() {
            return (value, targetType) -> fixed;
        }

        @Override
        public Notation.SchemaParser schemaParser() { return null; }
    };
}
```

**Problem:** Seven overridden methods, six of which return null. The one method that matters (`converter()`) is buried. Since Mockito is not in `ksml-data`'s test dependencies, a full implementation is required - but extracting it to a named inner class makes the intent clearer and keeps `notationConvertingTo` short.

**Suggested fix:**

```java
/** Minimal Notation stub that delegates conversion to a fixed return value. */
private static final class FixedResultNotation implements Notation {
    private final DataObject result;

    FixedResultNotation(DataObject result) { this.result = result; }

    @Override public Notation.SchemaUsage schemaUsage() { return null; }
    @Override public DataType defaultType() { return null; }
    @Override public String name() { return "stub"; }
    @Override public String filenameExtension() { return null; }
    @Override public Serde<Object> serde(DataType type, boolean isKey) { return null; }
    @Override public Notation.Converter converter() { return (value, targetType) -> result; }
    @Override public Notation.SchemaParser schemaParser() { return null; }
}

private static Notation notationConvertingTo(DataObject fixed) {
    return new FixedResultNotation(fixed);
}
```

**Reason:** The same pattern also appears in `NotationTest.minimalNotation()`. Named inner classes are reusable across tests, easier to search, and clarify which part of the behavior is under test versus which is just boilerplate.

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #652, opened 2026-07-06
Added unit tests for ksml-kafka-clients module

pantelis17:
# Added unit tests for `ksml-kafka-clients` module

## What this PR does

Adds a complete unit test suite for the `ksml-kafka-clients` module — the part of KSML that wraps Kafka clients to resolve tenant-prefixed topic names, group ids, and transactional ids transparently.

Before this PR, the module had no unit tests. Now it has **189 tests** across 35 test classes, covering every major class: admin client, consumer, producer, serializer/deserializer, resolvers, config, exceptions, and utilities.

It also includes two small production changes found while writing the tests: one bug fix and one rebalance-callback behavior fix.

---

### Bug fix 1 — `ResolvingClientConfig`

When a config map contained both a deprecated key (e.g. `topic.pattern`) and its replacement (`axual.topic.pattern`), the deprecated value was silently overwriting the current one. Fixed so the current key always wins.

### Bug fix 2 — `ResolvingConsumer` (rebalance-callback behavior change)

The proxy `ConsumerRebalanceListener` overrode `onPartitionsAssigned` and `onPartitionsRevoked` but not `onPartitionsLost`. Kafka's `ConsumerRebalanceListener` has a default `onPartitionsLost` that delegates to `onPartitionsRevoked`, so on partition loss the event was routed to the proxy's `onPartitionsRevoked`. The partition names were therefore *already* unresolved — but the **wrong callback fired**: a caller's `onPartitionsRevoked` ran on partition loss instead of their `onPartitionsLost`.

Added the missing `onPartitionsLost` override (unresolving partitions like the other two callbacks). Callers that implement both callbacks now correctly receive `onPartitionsLost` on loss.

**Behavior change:** on partition loss a consumer must not commit offsets for partitions it no longer owns, and Kafka Streams relies on the lost-vs-revoked distinction. Git blame should reflect the accurate reason.

### New test dependency — `pom.xml`

Added `mockito-junit-jupiter` as a test-scope dependency (which brings in `mockito-core` transitively). Mockito is used to mock Kafka's internal client classes (`KafkaAdminClient`, `KafkaConsumer`, `KafkaProducer`) so tests run without a real Kafka broker.

### New shared test double — `PrefixResolver`

A simple `TopicResolver` / `GroupResolver` test double in `io.axual.ksml.client.testutil` that resolves by prepending `tenant-` and unresolves by stripping it. Shared across all test packages.

### Admin client tests

| Test class | What it covers |
|---|---|
| `ResolvingAdminTest` | End-to-end: all admin operations resolve/unresolve topic and group names via the real `ResolvingAdmin` |
| `ResolvingCreateTopicsResultTest` | `values()` map is keyed by unresolved topic name |
| `ResolvingDeleteTopicsResultTest` | Same for delete |
| `ResolvingDescribeTopicsResultTest` | Same for describe |
| `ResolvingDescribeAclsResultTest` | ACL binding names are unresolved; failures propagate |
| `ResolvingDescribeConsumerGroupsResultTest` | Group descriptions keyed and named by unresolved group id |
| `ResolvingDeleteConsumerGroupsResultTest` | Unresolvable group ids are dropped |
| `ResolvingConsumerGroupOffsetsResultTest` | Partition resolution for delete/alter offset results |
| `ResolvingListTopicsResultTest` | Internal topics are kept; unresolvable entries are dropped |
| `ResolvingListConsumerGroupsResultTest` | `all()` and `valid()` return unresolved group ids |
| `ResolvingListOffsetsResultTest` | Offset results keyed by unresolved partition |
| `ExtendableListConsumerGroupsResultTest` | Base class delegation |
| `ResolverUtilTest` | ACL filter/binding resolve and unresolve helpers |

### Consumer tests

| Test class | What it covers |
|---|---|
| `ResolvingConsumerTest` | subscribe, assign, seek, commit, poll — all resolve/unresolve topics and partitions; rebalance listener unresolves assigned, revoked and lost partitions |
| `ForwardingConsumerTest` | Every `Consumer` method is forwarded to the delegate |
| `ResolvingConsumerConfigTest` | Group id and partition assignor are rewritten in the downstream config |
| `ResolvingConsumerPartitionAssignorTest` | Cluster and subscriptions are unresolved before backing assignor; assignments are resolved back |
| `ResolvingPartitionInfoTest` | Wraps topic name; leader/replica info throws `UnsupportedOperationException` |

### Producer tests

| Test class | What it covers |
|---|---|
| `ResolvingProducerTest` | send, sendOffsetsToTransaction, partitionsFor — all resolve/unresolve |
| `ForwardingProducerTest` | Every `Producer` method is forwarded to the delegate |
| `ResolvingProducerConfigTest` | Transactional id is resolved into the downstream config when a pattern is set |

### Resolver tests

| Test class | What it covers |
|---|---|
| `TopicResolverTest` | Pattern wrap, partition resolve/unresolve, collection and map helpers |
| `GroupResolverTest` | Collection resolve/unresolve, null handling |
| `TransactionalIdResolverTest` | Single and collection resolve/unresolve |
| `ResolvingClientConfigTest` | Instance/class-name/Class config lookup; deprecated key replacement |

### Other tests

| Test class | What it covers |
|---|---|
| `ResolvingSerializerTest` / `ResolvingDeserializerTest` | Topic name is resolved before delegating to backing serde |
| `ClientExceptionTest` / `ConfigExceptionTest` | Exception message format and inheritance |
| `FactoryUtilTest` | Class instantiation, missing/unknown/incompatible class errors |
| `MapUtilTest` | `toStringValues` and `merge` utilities |

---

## Test design notes

- No real Kafka broker needed — Mockito's `mockConstruction` intercepts `KafkaAdminClient`, `KafkaConsumer`, and `KafkaProducer` creation
- No internal Kafka APIs used in tests (`KafkaFutureImpl`, `CoordinatorKey`)
- All assertions use AssertJ dedicated methods
- All tests have `@DisplayName` for readable output
- Sonar: 0 open issues on the PR


ksml-kafka-clients/pom.xml
```diff
@@ -75,6 +75,16 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
```
KasparMetsa, line 80: ### 1. `pom.xml` declares `mockito-core` explicitly but `mockito-junit-jupiter` already pulls it in transitively

**PR code (`ksml-kafka-clients/pom.xml`):**

```xml
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <scope>test</scope>
</dependency>
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-junit-jupiter</artifactId>
    <scope>test</scope>
</dependency>
```

**Suggested fix:**

```xml
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-junit-jupiter</artifactId>
    <scope>test</scope>
</dependency>
```

**Reason:** `mockito-junit-jupiter` depends on `mockito-core` transitively. Declaring `mockito-core` separately is redundant - the version is already managed centrally via `<dependencyManagement>` in the root pom, and `mockito-junit-jupiter` will pull in the right version. This is the same pattern used in all other modules that use Mockito (e.g. `ksml-runner`).

ksml-kafka-clients/src/test/java/io/axual/ksml/client/resolving/GroupResolverTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Exercises the collection default methods of {@link GroupResolver} through a prefixing test double.
+ */
+class GroupResolverTest {
+    private static final String PREFIX = "resolved-";
+
+    private final GroupResolver resolver = new GroupResolver() {
```
KasparMetsa: ### `GroupResolverTest` and `TopicResolverTest` define their own anonymous inline resolver instead of reusing `PrefixResolver`

The PR explicitly introduces `PrefixResolver` as "a simple test double shared across all test packages". `PrefixResolver` implements both `TopicResolver` and `GroupResolver`. Yet both resolver tests ignore it and define the same logic inline:

**PR code (`GroupResolverTest.java`):**

```java
private final GroupResolver resolver = new GroupResolver() {
    @Override
    public String resolve(String name) {
        return name == null ? null : PREFIX + name;
    }

    @Override
    public String unresolve(String name) {
        if (name == null || !name.startsWith(PREFIX)) return null;
        return name.substring(PREFIX.length());
    }
};
```

The same anonymous class pattern is repeated verbatim in `TopicResolverTest.java`.

**Suggested fix** - use `PrefixResolver` in both:

```java
// GroupResolverTest.java
private final GroupResolver resolver = new PrefixResolver();
```

```java
// TopicResolverTest.java
private final TopicResolver resolver = new PrefixResolver();
```

Note: both tests use a local `PREFIX = "resolved-"` constant while `PrefixResolver.PREFIX` is `"tenant-"`. Switching requires updating expected values in assertions to use `PrefixResolver.PREFIX`.

**Reason:** The `PrefixResolver` was created specifically to be the shared test double for these tests. Using anonymous classes in the two classes it was built for contradicts that intent, adds noise, and means a future change to the prefix logic needs to be applied in three places instead of one. (`TransactionalIdResolverTest` is not affected - `PrefixResolver` does not implement `TransactionalIdResolver`, so its inline implementation is necessary.)

ksml-kafka-clients/src/test/java/io/axual/ksml/client/consumer/ResolvingConsumerTest.java
```diff
@@ -0,0 +1,361 @@
...
+                    revoked.addAll(partitions);
+                }
+
+                @Override
+                public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
+                    assigned.addAll(partitions);
+                }
+            };
+
+            consumer.subscribe(List.of(UNRESOLVED_TOPIC), callerListener);
```
KasparMetsa, line 108: ### `ResolvingConsumerTest.subscribeWithListenerWrapsCallbacks` does not cover `onPartitionsLost`

`ConsumerRebalanceListener` has three callback methods: `onPartitionsRevoked`, `onPartitionsAssigned`, and `onPartitionsLost`. The test only verifies that the wrapper forwards two of them:

**PR code (`ResolvingConsumerTest.java`):**

```java
final ConsumerRebalanceListener callerListener = new ConsumerRebalanceListener() {
    @Override
    public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
        revoked.addAll(partitions);
    }

    @Override
    public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
        assigned.addAll(partitions);
    }
};

consumer.subscribe(List.of(UNRESOLVED_TOPIC), callerListener);

final ArgumentCaptor<ConsumerRebalanceListener> wrapped = ArgumentCaptor.captor();
verify(delegate).subscribe(any(Collection.class), wrapped.capture());
wrapped.getValue().onPartitionsAssigned(List.of(RESOLVED_PARTITION));
wrapped.getValue().onPartitionsRevoked(List.of(RESOLVED_PARTITION));
```

`onPartitionsLost` is never invoked on the wrapped listener.

**Suggested fix** - add `onPartitionsLost` to the test:

```java
final var lost = new ArrayList<TopicPartition>();
final ConsumerRebalanceListener callerListener = new ConsumerRebalanceListener() {
    @Override
    public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
        revoked.addAll(partitions);
    }

    @Override
    public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
        assigned.addAll(partitions);
    }

    @Override
    public void onPartitionsLost(Collection<TopicPartition> partitions) {
        lost.addAll(partitions);
    }
};

// ... subscribe and capture wrapped ...

wrapped.getValue().onPartitionsAssigned(List.of(RESOLVED_PARTITION));
wrapped.getValue().onPartitionsRevoked(List.of(RESOLVED_PARTITION));
wrapped.getValue().onPartitionsLost(List.of(RESOLVED_PARTITION));

assertThat(assigned).containsExactly(UNRESOLVED_PARTITION);
assertThat(revoked).containsExactly(UNRESOLVED_PARTITION);
assertThat(lost).containsExactly(UNRESOLVED_PARTITION);
```

**Reason:** If `ResolvingConsumer` forgets to unresolve partitions in `onPartitionsLost`, or fails to forward the call to the caller's listener, there is no test to catch it. The default implementation of `onPartitionsLost` calls `onPartitionsRevoked`, so the behavior may already work by accident - but that is exactly the kind of thing a test should document and protect.

ksml-kafka-clients/src/test/java/io/axual/ksml/client/producer/ResolvingProducerTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+            final ArgumentCaptor<ProducerRecord<String, String>> sent = ArgumentCaptor.captor();
+            verify(delegate).send(sent.capture());
+            assertThat(sent.getValue().topic()).isEqualTo(RESOLVED_TOPIC);
+            assertThat(metadata.topic()).isEqualTo(UNRESOLVED_TOPIC);
+        });
+    }
+
+    @Test
+    @DisplayName("send with a null callback delegates to the single-argument send")
```
KasparMetsa, line 78: ### `sendWithNullCallbackDelegates` - display name claims more than the test verifies

**PR code (`ResolvingProducerTest.java`):**

```java
@Test
@DisplayName("send with a null callback delegates to the single-argument send")
void sendWithNullCallbackDelegates() throws Exception {
    withProducer((producer, delegate) -> {
        when(delegate.send(any())).thenReturn(resolvedMetadataFuture());

        producer.send(new ProducerRecord<>(UNRESOLVED_TOPIC, "k", "v"), null).get();

        verify(delegate).send(any());
    });
}
```

**Problem:** `verify(delegate).send(any())` only verifies that the single-argument `send` was called on the delegate. It does not verify that the null callback was the reason. If the production code calls `delegate.send(record, null)` instead of `delegate.send(record)`, the test would fail - but the display name says "delegates to the single-argument send", which is an implementation detail of how `null` callbacks are handled. If the production code changes to pass `null` through instead, the test breaks even though the behavior (not calling back) is unchanged.

**Suggested fix** - either:

```java
// Option A: test the observable behaviour, not the implementation detail
@DisplayName("send with a null callback does not invoke any callback")
void sendWithNullCallbackDoesNotInvokeCallback() throws Exception { ... }
```

Or:

```java
// Option B: if the single-arg delegation is intentional, verify it explicitly
verify(delegate, never()).send(any(), any());  // two-arg was NOT called
verify(delegate).send(any());                  // single-arg WAS called
```

**Reason:** The display name and the test body are slightly misaligned. The test only confirms the single-arg overload is called; it doesn't verify the null callback has no effect (which is the user-visible behaviour). Clarifying either the name or the assertion makes the intent unambiguous.

KasparMetsa (changes requested):

ksml-kafka-clients/src/main/java/io/axual/ksml/client/consumer/ResolvingConsumer.java
```diff
@@ -365,6 +365,11 @@ public void onPartitionsRevoked(Collection<TopicPartition> collection) {
         public void onPartitionsAssigned(Collection<TopicPartition> collection) {
```
KasparMetsa, line 365: The `onPartitionsLost` change is described wrongly, and it is an unannounced behavior change

**Severity: Low** (the code is correct; only the description and change-communication are wrong)

**Code in PR** (`ResolvingConsumer.java`) - the code itself is correct:

```java
@Override
public void onPartitionsAssigned(Collection<TopicPartition> collection) {
    listener.onPartitionsAssigned(topicResolver.unresolveTopicPartitions(collection));
}

@Override
public void onPartitionsLost(Collection<TopicPartition> collection) {
    listener.onPartitionsLost(topicResolver.unresolveTopicPartitions(collection));
}
```

The PR body says:

> ...a caller listening for lost partitions received the resolved (tenant-prefixed) names instead of
> the original ones.

That is not what happened. Kafka's interface has a default:

```java
default void onPartitionsLost(Collection<TopicPartition> partitions) {
    onPartitionsRevoked(partitions);   // calls THIS proxy's onPartitionsRevoked
}
```

So before the fix, when the delegate consumer lost partitions it called the proxy's default
`onPartitionsLost`, which called the proxy's own `onPartitionsRevoked`, which **did** unresolve the
names and delivered them to the caller's `onPartitionsRevoked`. In other words:

- The names were **already unresolved** (the "resolved names" claim is wrong).
- The real bug was the **wrong callback fired**: the caller's `onPartitionsRevoked` ran instead of
  their `onPartitionsLost`.

**Suggested change:** keep the code (it is correct), but fix the description and flag it as a behavior
change:

> Before this fix the proxy did not override `onPartitionsLost`, so Kafka's default routed lost
> partitions to `onPartitionsRevoked`. Callers that implement both callbacks now correctly receive
> `onPartitionsLost` on partition loss.

**Why Low:** nothing needs to change in the code, only the PR text - so it is not a blocker. It is
ranked above the other Low items because the behavior change it introduces is meaningful: on partition
loss a consumer must **not** commit offsets for partitions it no longer owns, and Kafka Streams relies
on the `lost` vs `revoked` distinction. Future readers of `git blame` need the accurate reason.

ksml-kafka-clients/src/main/java/io/axual/ksml/client/resolving/ResolvingClientConfig.java
```diff
@@ -142,8 +142,10 @@ public static <T> void replaceDeprecatedConfigKeys(Map<String, T> config) {
     private static <T> void replaceDeprecatedConfigKeys(final Map<String, T> config, String deprecatedKey, String currentKey) {
```
KasparMetsa, line 142: **Code in PR** (`ResolvingClientConfig.java`):

```java
private static <T> void replaceDeprecatedConfigKeys(final Map<String, T> config, String deprecatedKey, String currentKey) {
    if (config.containsKey(deprecatedKey)) {
        log.warn("The configuration key '{}' is deprecated. Use the configuration key '{}' instead.", deprecatedKey, currentKey);
        final T value = config.remove(deprecatedKey);
        if (!config.containsKey(currentKey)) {
            config.put(currentKey, value);
        }
    }
}
```

When **both** keys are present, the code now (correctly) keeps the current key and drops the
deprecated one - but it still logs "is deprecated, use X instead" and **silently discards** the
deprecated value. The operator is told to switch to a key they are already using, with no hint that
their deprecated value was ignored.

**Suggested change:**

```java
private static <T> void replaceDeprecatedConfigKeys(final Map<String, T> config, String deprecatedKey, String currentKey) {
    if (!config.containsKey(deprecatedKey)) {
        return;
    }
    final T value = config.remove(deprecatedKey);
    if (config.containsKey(currentKey)) {
        log.warn("Both the deprecated key '{}' and the current key '{}' are set. Ignoring the deprecated value.", deprecatedKey, currentKey);
    } else {
        log.warn("The configuration key '{}' is deprecated. Use the configuration key '{}' instead.", deprecatedKey, currentKey);
        config.put(currentKey, value);
    }
}
```

**Why Low:** also a silent behavior change (previously the deprecated value overwrote the current one),
but the fix direction is right, so this is diagnostics polish - a wrong deprecated value that is now
silently ignored is just harder to debug without a clear log.

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #654, opened 2026-07-08
Added tests for ksml module

pantelis17:
# Add unit tests for the `ksml` module

## What this PR does

This PR adds many unit tests for the `ksml` module. Before, this module had
very little test coverage. Now the main building blocks are tested.

In short (**107 files changed**, ~8,000 lines):

- **102 new test files**
- **4 shared test helper files** to avoid copy-paste
- **1 tiny production bug fix**

## Why

The `ksml` module had almost no tests. That makes changes risky, because a
small mistake can break things without anyone noticing. These tests give us a
safety net and document how each part is meant to work.

## What is covered

Tests were added for the main areas of the module:

- **Operations** – map, filter, join, count, reduce, aggregate, branch,
  convert, transform, window, suppress, and more
- **Parsers** – operation, join, `to`, and definition parsers
- **Processors** – peek, transform key/value, metadata
- **User functions** – predicates, joiners, transformers, extractors, etc.
- **Store proxies** – key-value, window, session, versioned, timestamped
- **Metrics, execution, exceptions, schema parsers, and data mappers**

## How the tests are written

The tests follow a few simple rules:

- They **check real behavior**, not only "no error happened". For example, a
  join test mocks the Kafka Streams object and checks that the correct
  `join` / `leftJoin` / `outerJoin` method is really called with the right
  arguments.
- They use **AssertJ** for readable assertions.
- Repeated cases use **parameterized tests** instead of copy-paste.
- Common setup is moved into small helper classes
  (`OperationTestSupport`, `UserTestSupport`, `OperationParserTestSupport`,
  `SchemaParserTestSupport`) so each test stays short and clear.

## Production change

One small bug was found and fixed while writing the tests:

- `StructSchemaFieldParser` built its field list with `Arrays.asList(...)`,
  which returns a **fixed-size** list. A later `removeIf(...)` call would fail
  with `UnsupportedOperationException`. It now uses a normal `ArrayList`, so
  removing items works.

## Result

- All tests pass: **738 run, 0 failures, 0 errors**
- No behavior change for users (only the small parser fix above)


ksml/src/test/java/io/axual/ksml/operation/FilterOperationTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+class FilterOperationTest {
+
+    private FilterOperation operation() {
+        return new FilterOperation(storeConfig("filter"), predicate());
+    }
+
+    @Test
+    void applyToStreamReturnsStream() {
+        assertThat(operation().apply(kStream(), mockContext())).isInstanceOf(KStreamWrapper.class);
```
KasparMetsa: ### 1. `FilterOperationTest` and `FilterNotOperationTest` are structurally identical - neither verifies which stream method is called

`FilterOperation` and `FilterNotOperation` should call `.filter()` and `.filterNot()` respectively on the underlying `KStream` or `KTable`. Both test classes are byte-for-byte identical after normalising the class names - neither test checks which method is actually called.

**PR code (`FilterOperationTest.java` and `FilterNotOperationTest.java`):**

```java
@Test
void applyToStreamReturnsStream() {
    assertThat(operation().apply(kStream(), mockContext())).isInstanceOf(KStreamWrapper.class);
}

@Test
void applyToTableReturnsTable() {
    assertThat(operation().apply(kTable(), mockContext())).isInstanceOf(KTableWrapper.class);
}
```

Both files have this exact structure. A bug where `FilterNotOperation` calls `.filter()` instead of `.filterNot()` passes all tests because neither class verifies the actual Kafka Streams method called.

The same issue applies to any pair of operations that wrap different Kafka Streams methods but share the same input/output type: `CountOperation` never verifies `.count()` was called; `AggregateOperation` never verifies `.aggregate()`; `ReduceOperation` never verifies `.reduce()`.

**Suggested fix** - create the mock stream explicitly and verify the correct method:

```java
// FilterOperationTest.java
@Test
@SuppressWarnings("unchecked")
void applyToStreamCallsFilter() {
    final KStream<Object, Object> stream = mock(KStream.class);
    final var input = new KStreamWrapper(stream, key(), value());

    operation().apply(input, mockContext());

    verify(stream).filter(any(), any(Named.class));  // not filterNot
}

// FilterNotOperationTest.java
@Test
@SuppressWarnings("unchecked")
void applyToStreamCallsFilterNot() {
    final KStream<Object, Object> stream = mock(KStream.class);
    final var input = new KStreamWrapper(stream, key(), value());

    operation().apply(input, mockContext());

    verify(stream).filterNot(any(), any(Named.class));  // not filter
}
```

**Reason:** A test that only checks the return type (`isInstanceOf(KStreamWrapper.class)`) passes even if the operation calls the completely wrong Kafka Streams method. The point of unit-testing an operation is to confirm it delegates to the right method. Several other operation tests in this PR already do this correctly (e.g. `JoinWithStreamOperationTest` verifies `stream.join(...)`, `GroupByOperationTest` verifies `stream.groupBy(...)`).

ksml/src/test/java/io/axual/ksml/operation/parser/OperationParserTestSupport.java
```diff
@@ -0,0 +1,52 @@
...
+
+    static void registerNotations() {
+        final var jsonNotation = new JsonNotation(new NotationContext(new DataObjectFlattener(), new DataTypeFlattener()));
+        ExecutionContext.INSTANCE.notationLibrary().register(UserType.DEFAULT_NOTATION,
+                new BinaryNotation(new NotationContext(new DataObjectFlattener(), new DataTypeFlattener()), jsonNotation::serde));
+        ExecutionContext.INSTANCE.notationLibrary().register(JsonNotation.NOTATION_NAME, jsonNotation);
+    }
+
+    static ParseNode nodeOf(String yaml) throws Exception {
+        return ParseNode.fromRoot(YAMLObjectMapper.INSTANCE.readValue(yaml, JsonNode.class), "test");
```
KasparMetsa: ### 2. `nodeOf()` is duplicated between `OperationParserTestSupport` and `SchemaParserTestSupport`

Both helpers define an identical `static ParseNode nodeOf(String yaml)` method.

**PR code (`OperationParserTestSupport.java`):**

```java
static ParseNode nodeOf(String yaml) throws Exception {
    return ParseNode.fromRoot(YAMLObjectMapper.INSTANCE.readValue(yaml, JsonNode.class), "test");
}
```

**PR code (`SchemaParserTestSupport.java`):**

```java
static ParseNode nodeOf(String yaml) throws Exception {
    final var root = YAMLObjectMapper.INSTANCE.readValue(yaml, JsonNode.class);
    return ParseNode.fromRoot(root, "test");
}
```

Same implementation, different local variable style.

**Suggested fix** - move `nodeOf` into a single shared utility (e.g. as a method on `OperationParserTestSupport`, or into a new package-independent `TestParseSupport`), and have `SchemaParserTestSupport` delegate to it or be removed:

```java
// One canonical location - e.g. OperationParserTestSupport or a new TestParseSupport
static ParseNode nodeOf(String yaml) throws Exception {
    return ParseNode.fromRoot(YAMLObjectMapper.INSTANCE.readValue(yaml, JsonNode.class), "test");
}
```

**Reason:** The `OperationParserTestSupport` class was added to avoid copy-paste between tests. Having two identical implementations of the same utility in two different helpers defeats that purpose. If the YAML parsing logic changes, it needs to be updated in two places.

ksml/src/test/java/io/axual/ksml/operation/CountOperationTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.mockito.Mockito.verify;
+
+class CountOperationTest {
+
+    private CountOperation operation() {
+        return new CountOperation(storeConfig("count"));
+    }
+
+    @Test
+    void applyToGroupedStreamReturnsTable() {
```
KasparMetsa: ### 3. `@DisplayName` is absent on 334 out of 341 tests

Every other recent test PR in this repo (636, 644, 646, 652) uses `@DisplayName` on all tests. This PR uses it on 7 tests only, all in `HeaderDataObjectMapperTest`. The remaining 334 tests identify themselves only by method name.

**PR code (typical pattern across 90+ test files):**

```java
@Test
void applyToGroupedStreamReturnsTable() {
    assertThat(operation().apply(groupedStream(), mockContext())).isInstanceOf(KTableWrapper.class);
}
```

**Suggested fix:**

```java
@Test
@DisplayName("apply to a grouped stream produces a KTable result")
void applyToGroupedStreamReturnsTable() {
    assertThat(operation().apply(groupedStream(), mockContext())).isInstanceOf(KTableWrapper.class);
}
```

**Reason:** Without `@DisplayName`, the test report shows raw method names like `applyToGroupedStreamReturnsTable`. With it, parameterized tests and hierarchical reports are much more readable. More importantly, `@DisplayName` documents intent in a way that survives renaming - several camelCase method names in this PR are borderline cryptic when read in a CI report.

ksml/src/test/java/io/axual/ksml/operation/FilterOperationTest.java
pantelis17: Fixed Filter/FilterNot to verify the table path calls the right method (verify(table).filter(...) + never() on the sibling), and did the same for Count/Aggregate/Reduce on the grouped paths.

The verify(stream).filter(...) suggestion won't work though: on a stream both go through processValues(...) with a different processor, so filter/filterNot is never called there. Left that path as-is, let me know if you want me to assert processValues instead.

ksml/src/test/java/io/axual/ksml/operation/OperationTestSupport.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    /**
+     * Registers a mock default notation returning a mock Serde. Operations that build
+     * Grouped/Produced/Joined/Repartitioned instances call {@link StreamDataType#serde()}, which
+     * resolves a notation from the global (shared) library. Because other test classes overwrite
+     * that global entry, this is re-registered by {@link #mockContext()} before every {@code apply}.
+     */
+    @SuppressWarnings("unchecked")
+    private static void registerMockDefaultNotation() {
```
KasparMetsa, line 117: **Code in PR** (`OperationTestSupport.java`):

```java
private static void registerMockDefaultNotation() {
    final var notation = mock(Notation.class);
    final Serde<Object> serde = mock(Serde.class);
    when(notation.serde(any(), anyBoolean())).thenReturn(serde);
    ExecutionContext.INSTANCE.notationLibrary().register(UserType.DEFAULT_NOTATION, notation);
}
// ...
static TopologyBuildContext mockContext() {
    registerMockDefaultNotation();   // called on every apply()
    ...
}
```

The helper writes a mock into `ExecutionContext.INSTANCE` - a process-wide singleton shared with the
rest of the module's tests. The Javadoc even admits the problem:

> Because other test classes overwrite that global entry, this is re-registered by `mockContext()`
> before every `apply`.

It works today only because Surefire runs sequentially and each test re-registers before use. But:

- After the operation tests run, a mock notation is left registered in the shared library.
- It couples these tests to the module's real topology/notation tests through global state - order
  now matters.
- If the module ever enables parallel test execution (`forkCount`/`parallel`), tests on different
  threads will race on the shared singleton.

**Suggested change:** save and restore the previous entry so no state leaks:

```java
@BeforeEach
void saveNotation() { previous = ExecutionContext.INSTANCE.notationLibrary().get(UserType.DEFAULT_NOTATION); }

@AfterEach
void restoreNotation() { ExecutionContext.INSTANCE.notationLibrary().register(UserType.DEFAULT_NOTATION, previous); }
```

(or expose a reset hook on the notation library). If restoring is not practical because of the
production singleton, document explicitly that parallel execution must stay off for this module.

**Why first:** it is the only item that can affect build reliability. The current design is one config
change away from order-dependent, flaky failures, and it silently couples unrelated test classes.

ksml/src/test/java/io/axual/ksml/operation/AggregateOperationTest.java
```diff
@@ -0,0 +1,150 @@
...
+    void applyToGroupedTableCallsAggregate() {
+        final KGroupedTable<Object, Object> grouped = mock(KGroupedTable.class);
+        final var input = new KGroupedTableWrapper(grouped, key(), value());
+
+        assertThat(operation().apply(input, mockContext())).isInstanceOf(KTableWrapper.class);
+        verify(grouped).aggregate(any(Initializer.class), any(Aggregator.class), any(Aggregator.class), any(Named.class));
+    }
+
+    @Test
+    @DisplayName("applying to a session-windowed stream returns a KTable")
```
KasparMetsa: **Code in PR** (`AggregateOperationTest.java`; same shape in `CountOperationTest`, `ReduceOperationTest`,
and two methods of `SuppressOperationTest`):

```java
@Test
@DisplayName("applying to a session-windowed stream returns a KTable")
void applyToSessionWindowedReturnsTable() {
    assertThat(operation().apply(sessionWindowed(), mockContext())).isInstanceOf(KTableWrapper.class);
}
```

This passes as long as *some* `KTableWrapper` comes back - it would still pass if the operation
delegated to `count()` instead of `aggregate()` on the windowed stream. The **grouped** cases in the
same class already show the right pattern:

```java
void applyToGroupedStreamCallsAggregate() {
    final KGroupedStream<Object, Object> grouped = mock(KGroupedStream.class);
    final var input = new KGroupedStreamWrapper(grouped, key(), value());
    assertThat(operation().apply(input, mockContext())).isInstanceOf(KTableWrapper.class);
    verify(grouped).aggregate(any(Initializer.class), any(Aggregator.class));   // <-- verifies behavior
}
```

**Suggested change:** build the windowed/cogrouped mock in the test (as the grouped tests do) and
verify the specific method:

```java
void applyToSessionWindowedCallsAggregate() {
    final SessionWindowedKStream<Object, Object> windowed = mock(SessionWindowedKStream.class);
    final var input = new SessionWindowedKStreamWrapper(windowed, key(), value());
    assertThat(operation().apply(input, mockContext())).isInstanceOf(KTableWrapper.class);
    verify(windowed).aggregate(any(Initializer.class), any(Aggregator.class), any(Merger.class), any(Materialized.class));
}
```

**Why:** these windowed delegations are exactly the higher-risk ones (each windowed type has its own
`aggregate`/`count`/`reduce`), and asserting only the return type is inconsistent with the grouped
tests in the very same class.

**Note - the stream-path `applyToStreamReturnsStream` tests (Filter, Peek, Transform*) are a different
case:** those genuinely cannot verify a distinct method (filter and filterNot both delegate to
`KStream.processValues`, differing only in the processor), and the code says so in a comment. Their
real behavior is covered by the separate `*ProcessorTest` classes. That split is acceptable; only the
windowed case above is a real gap.

ksml/src/test/java/io/axual/ksml/operation/FilterOperationTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static io.axual.ksml.operation.OperationTestSupport.predicate;
+import static io.axual.ksml.operation.OperationTestSupport.storeConfig;
+import static io.axual.ksml.operation.OperationTestSupport.value;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+class FilterOperationTest {
```
KasparMetsa: The new mock-based operation tests overlap with, and diverge from, the existing `@KSMLTest` topology tests

**Severity: Medium**

The module already tests operations end-to-end with real Kafka Streams behavior:

```java
// existing convention - KSMLFilterTest.java
@ExtendWith({KSMLTestExtension.class})
@KSMLTest(topology = "pipelines/test-filter.yaml", schemaDirectory = "schemas")
void testFilterAvroRecords() { ... asserts real filtered output ... }
```

`KSMLFilterTest`, `KSMLFilterNotTest`, `KSMLJoinTest`, `KSMLOuterJoinTest`, `KSMLCountTest`,
`KSMLReduceTest`, `KSMLBranchingTest`, `KSMLTransformKey*Test`, `KSMLMergeTest`,
`KSMLRepartitionTest`, etc. already exist. This PR adds a **parallel** set of mock-based
`FilterOperationTest`, `JoinWith*OperationTest`, `CountOperationTest`, `ReduceOperationTest`,
`BranchOperationTest`, `TransformKey*OperationTest`, `MergeOperationTest`, `RepartitionOperationTest`,
... that mostly assert "the right `KStream` method was called" or (P2) the wrapper type.

### Specific overlaps (operation already has a topology test AND gets a new mock test)

| Operation | Existing topology test (origin/main) | New mock test (this PR) |
|---|---|---|
| filter | `KSMLFilterTest` | `FilterOperationTest` |
| filterNot | `KSMLFilterNotTest` | `FilterNotOperationTest` |
| join (stream-stream / stream-table) | `KSMLJoinTest` | `JoinWithStreamOperationTest`, `JoinWithTableOperationTest` |
| outerJoin | `KSMLOuterJoinTest` | `OuterJoinWithStreamOperationTest`, `OuterJoinWithTableOperationTest` |
| count | `KSMLCountTest` | `CountOperationTest` |
| reduce | `KSMLReduceTest` | `ReduceOperationTest` |
| aggregate / cogroup | `KSMLAggregatePipelinesTest`, `KSMLAggregateStoreTest` | `AggregateOperationTest`, `CogroupOperationTest` |
| branch | `KSMLBranchingTest` | `BranchOperationTest` |
| groupBy | `KSMLGroupByTest` | `GroupByOperationTest` |
| merge | `KSMLMergeTest` | `MergeOperationTest` |
| repartition | `KSMLRepartitionTest` | `RepartitionOperationTest` |
| transformKey | `KSMLTransformKeyTest` | `TransformKeyOperationTest` |
| transformKeyValue | `KSMLTransformKeyValueTest` | `TransformKeyValueOperationTest` |
| transformKeyValueToKeyValueList | `KSMLTransformKeyValueToKeyValueListTest` | `TransformKeyValueToKeyValueListOperationTest` |
| transformKeyValueToValueList | `KSMLTransformKeyValueToValueListTest` | `TransformKeyValueToValueListOperationTest` |

That is ~15 operations tested twice, at two fidelity levels.

### Side-by-side: the existing test checks real output; the new test checks a method call

**filter** - `KSMLFilterTest` pipes records with different colors and checks the actual filtered
output:

```java
// KSMLFilterTest (existing, topology)
inputTopic.pipeInput("key", sensorData.toRecord());     // several colors
assertFalse(outputTopic.isEmpty());
assertEquals(2, outputValues.size());                   // only the "blue" ones survive
```

```java
// FilterOperationTest (new, mock)
verify(table).filter(any(Predicate.class), any(Named.class));   // stream case only checks wrapper type
```

**join** - `KSMLJoinTest` pipes alert settings + sensor data and reads the joined result:

```java
// KSMLJoinTest (existing, topology)
alertSettings.pipeInput("Amsterdam", ...); sensorIn.pipeInput("sensor1", ...);
assertEquals(2, sensorAlerts.getQueueSize());
List<String> valuesList = sensorAlerts.readValuesToList();   // asserts the real joined values
```

```java
// JoinWithStreamOperationTest (new, mock)
verify(stream).join(any(KStream.class), any(ValueJoinerWithKey.class), any(JoinWindows.class), any(StreamJoined.class));
```

**count** - `KSMLCountTest` pipes 6 records over 3 keys and checks the counts:

```java
// KSMLCountTest (existing, topology)
inputTopic.pipeInput("keyFirst", "value1"); ... // 6 inputs
assertEquals(3, keyValueMap.get("keyFirst"));   // real counting result
```

```java
// CountOperationTest (new, mock)
verify(grouped).count(any(Named.class));        // windowed cases only check wrapper type (see P2)
```

In each pair the existing test proves the operation *does the right thing*; the new test only proves
*which Streams method is wired up*.

### For balance - where the new tests genuinely add coverage (no topology counterpart)

These have no existing `KSML*Test` and are the valuable part of the PR: `AsOperationTest`,
`ConvertKey/Value/KeyValueOperationTest`, `ForEachOperationTest`, `PeekOperationTest`,
`PrintOperationTest`, `SuppressOperationTest`, `ToStream/ToTable/ToTopicNameExtractorOperationTest`,
`JoinWithGlobalTable`/`LeftJoinWithGlobalTableOperationTest`, `LeftJoinWith*OperationTest`,
`WindowBySession/WindowByTimeOperationTest`, `TransformMetadata/TransformValueOperationTest`,
`StreamOperationTest`, `BaseOperationTest`.

And these existing topology tests got **no** new mock test (fine - just showing the split is
uneven): `KSMLCopyTest`, `KSMLMapTest`, `KSMLMapKeyTest`, `KSMLMapValueAndAliasesTest`,
`KSMLFlatMapTest`, `KSMLFlatMapValuesTest`, `KSMLSelectKeyTest`, `KSMLRoutingTest`,
`KSMLIteratorTest`, `KSMLStateStoreTest`, `KSMLVersionedStateStoreTest`,
`KSMLStreamPartitionerTest`, `KSMLFieldModificationTest`, `KSMLTupleListTypeTest`,
`KSMLTransformToNoneTest`.

**Suggested change:** be explicit about the split of responsibility, and prefer the higher-fidelity
layer for behavior:

- Keep the unit tests for what the topology tests cannot easily reach (type deduction, store
  materialization wiring, error paths, parser edge cases).
- For "does this operation actually do X", lean on / extend the existing `@KSMLTest` tests.
- Add a short note in the PR (or a package `README`/`package-info`) explaining why both layers exist,
  so the next contributor knows which to add to.

**Why:** two parallel testing conventions for the same code raises maintenance cost and drifts over
time. Tests that mirror the implementation call-for-call (verify `stream.join(...)`) are
change-detector tests: they break on harmless refactors without necessarily catching real bugs, which
the topology tests already guard against.

ksml/src/test/java/io/axual/ksml/operation/OuterJoinWithStreamOperationTest.java
```diff
@@ -0,0 +1,58 @@
...
+import static io.axual.ksml.operation.OperationTestSupport.valueJoiner;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+class OuterJoinWithStreamOperationTest {
+
+    @Test
+    @DisplayName("outer join with a stream produces a KStream result using a JoinWindows-based outerJoin")
```
KasparMetsa, line 47: Heavy structural duplication across operation test classes

**Severity: Low-Medium**

Whole families of test classes are near-identical, differing only in the operation name and the
verified method (they *do* verify the correct method, so this is duplication, not false confidence).
The affected families total ~800 lines across ~13 classes:

| Family | Classes (lines each) |
|---|---|
| stream join | `JoinWithStreamOperationTest` (82), `LeftJoinWithStreamOperationTest` (58), `OuterJoinWithStreamOperationTest` (58) |
| table join | `JoinWithTableOperationTest` (109), `LeftJoinWithTableOperationTest` (109), `OuterJoinWithTableOperationTest` (54) |
| global-table join | `JoinWithGlobalTableOperationTest` (58), `LeftJoinWithGlobalTableOperationTest` (58) |
| convert | `ConvertKeyOperationTest` (52), `ConvertValueOperationTest` (53), `ConvertKeyValueOperationTest` (52) |
| transform | `TransformKeyOperationTest` (53), `TransformKeyValueOperationTest` (53), `TransformMetadataOperationTest` (53), `TransformValueOperationTest` (65), `TransformKeyValueToKeyValueListOperationTest` (53), `TransformKeyValueToValueListOperationTest` (53) |

**Exact example 1 - `LeftJoinWithStreamOperationTest` vs `OuterJoinWithStreamOperationTest`** are 58
identical lines except for **5**. The complete `diff` between the two files:

```diff
< import org.apache.kafka.streams.kstream.ValueJoinerWithKey;
> import org.apache.kafka.streams.kstream.ValueJoiner;
< class LeftJoinWithStreamOperationTest {
> class OuterJoinWithStreamOperationTest {
<     @DisplayName("left-joining two streams over a window delegates to KStream.leftJoin and returns a stream")
>     @DisplayName("outer join with a stream produces a KStream result using a JoinWindows-based outerJoin")
<         final var operation = new LeftJoinWithStreamOperation(dualStoreConfig("leftJoin"), streamDefinition(), valueJoiner(), Duration.ofSeconds(1), null);
>         final var operation = new OuterJoinWithStreamOperation(dualStoreConfig("outerJoin"), streamDefinition(), valueJoiner(), Duration.ofSeconds(1), null);
<         verify(stream).leftJoin(any(KStream.class), any(ValueJoinerWithKey.class), any(JoinWindows.class), any(StreamJoined.class));
>         verify(stream).outerJoin(any(KStream.class), any(ValueJoiner.class), any(JoinWindows.class), any(StreamJoined.class));
```

**Exact example 2 - `JoinWithTableOperationTest` vs `LeftJoinWithTableOperationTest`** are 109
identical lines; the entire difference is a mechanical `join` -> `leftJoin` across 4 test methods:

```diff
< class JoinWithTableOperationTest {
> class LeftJoinWithTableOperationTest {
<         verify(stream).join(any(KTable.class), any(ValueJoinerWithKey.class), any(Joined.class));
>         verify(stream).leftJoin(any(KTable.class), any(ValueJoinerWithKey.class), any(Joined.class));
<         verify(table).join(any(KTable.class), any(Function.class), any(ValueJoiner.class), any(TableJoined.class));
>         verify(table).leftJoin(any(KTable.class), any(Function.class), any(ValueJoiner.class), any(TableJoined.class));
<         verify(table).join(any(KTable.class), any(Function.class), any(ValueJoiner.class), any(TableJoined.class), any(Materialized.class));
>         verify(table).leftJoin(any(KTable.class), any(Function.class), any(ValueJoiner.class), any(TableJoined.class), any(Materialized.class));
<         verify(table).join(any(KTable.class), any(ValueJoiner.class), any(Named.class), any(Materialized.class));
>         verify(table).leftJoin(any(KTable.class), any(ValueJoiner.class), any(Named.class), any(Materialized.class));
```

(plus the class name and the `@DisplayName` / constructor `Join` -> `LeftJoin` renames.)

**Exact example 3 - `ConvertKeyOperationTest` vs `ConvertValueOperationTest`** (52-53 lines) differ
only by:

```diff
> import org.apache.kafka.streams.kstream.ValueMapper;   // extra import in the Value variant
< class ConvertKeyOperationTest {                          void applyToStreamSelectsKey()
> class ConvertValueOperationTest {                        void applyToStreamMapsValues()
<         final var operation = new ConvertKeyOperation(operationConfig("convertKey"), UNKNOWN_TYPE);
>         final var operation = new ConvertValueOperation(operationConfig("convertValue"), UNKNOWN_TYPE);
<         verify(stream).selectKey(any(), any(Named.class));
>         verify(stream).mapValues(any(ValueMapper.class), any(Named.class));
```

**Exact example 4 - the transform family** all share the same body; only the imported supplier type,
the factory helper, and the verified method change, e.g. `TransformKeyOperationTest` vs
`TransformMetadataOperationTest`:

```diff
<         verify(stream).process(any(ProcessorSupplier.class), any(Named.class));
>         verify(stream).processValues(any(FixedKeyProcessorSupplier.class), any(Named.class), any(String[].class));
```

**Suggested change:** collapse each family into one parameterized test:

```java
@ParameterizedTest
@MethodSource("streamJoinCases")
void streamJoinDelegatesToTheRightMethod(BaseOperation op, Consumer<KStream<Object,Object>> verifyCall) { ... }
```

**Why:** ~15+ classes repeat the same setup. A change to `apply(...)` or the join signature means
editing all of them by hand. The PR already parameterizes some cases, so this just applies that
approach consistently.

KasparMetsa (changes requested):

ksml/src/test/java/io/axual/ksml/operation/FilterOperationTest.java
KasparMetsa: I agree, that's right that `verify(stream).filter(...)` cannot work.

About the "leave as-is / assert processValues instead" part:

Asserting processValues instead would not really help, because both operations call it in the exact same way. The stream tests for filter and filterNot would still be identical and neither would prove which operation is which.

The part that actually differs is which processor gets wired in, and that one is easy to check, since FixedKeyOperationProcessorSupplier.get() returns the processor directly:
```
var captor = ArgumentCaptor.<FixedKeyProcessorSupplier<Object, Object, Object>>captor();
verify(stream).processValues(captor.capture(), any(Named.class), any());
assertThat(captor.getValue().get()).isInstanceOf(FilterProcessor.class); // FilterNotProcessor in the sibling test
```
This is safe (the predicate is only stored, not run), and it makes the two stream tests test genuinely different behavior.

Small related note: the PR adds FilterNotProcessorTest but not FilterProcessorTest, so FilterProcessor is only covered end to end by KSMLFilterTest. The captor check above (or a small FilterProcessorTest) would also close that gap.

If you'd rather keep the operation tests focused on wiring only and lean on the topology tests for stream behavior, that's fine too - could you just add a short comment saying so, and maybe add the missing FilterProcessorTest for symmetry?

ksml/src/test/java/io/axual/ksml/operation/OperationTestSupport.java
pantelis17: Applied the save/restore you suggested, but as an auto-registered JUnit extension (DefaultNotationIsolationExtension via META-INF/services + junit-platform.properties) rather than per-class @BeforeEach/@AfterEach, since the mock is registered by a shared helper (OperationTestSupport.mockContext()) used by ~38 classes — one extension isolates all of them without duplicating the hook. Added NotationLibrary.unregister() so the previously-absent entry is restored exactly, and kept parallel execution off (documented on the extension).

ksml/src/test/java/io/axual/ksml/operation/OuterJoinWithStreamOperationTest.java
pantelis17: Fair point on the duplication — these families are genuinely near-identical. I've opted to keep the explicit per-operation classes for now, for two reasons:

1. Readability/localization: each class is ~12 lines and reads at a glance; a failure names the exact operation. A parameterized version carrying an operation + a verify-lambda per case is denser and harder to scan, and for the table-join family (4 distinct apply scenarios each) a 2-D parameterization would read worse than what's there.

2. Low, mechanical churn: it's "boring" duplication — the only realistic trigger for a mass edit is a Kafka Streams signature change, which is rare and mechanical.

Since this is Low-Medium and about duplication rather than test confidence (the methods are correctly verified), I'd prefer to leave it. Happy to revisit as a follow-up if the families keep growing — the single-method ones (convert, transform) would parameterize cleanly if we decide it's worth it.

ksml/src/test/java/io/axual/ksml/operation/FilterOperationTest.java
pantelis17: Agreed on making the split explicit — added a package-info.java to the io.axual.ksml.operation test package documenting the two layers and which one to extend:

 - Unit tests (this package): fast, JVM-only, no topology/GraalVM runtime. They own type deduction, store-materialization wiring, error paths, parser edge cases, and "delegates to the correct Streams method" (join vs leftJoin).
 - @KSMLTest topology tests: own real end-to-end behavior — "does operation X actually produce Y".

On keeping both layers rather than folding the overlap into the topology tests: the unit layer runs without the Python/GraalVM runtime and the test-driver, and localizes a failure to a single operation, so it's a useful fast tier under the topology tests — not a replacement. I've kept the wiring assertions but the package doc now makes clear they are the first line of defence and intentionally don't re-prove behavior the topology tests own. The earlier P2 point (windowed cases only asserting wrapper type) is fixed in this PR — those now verify the specific method. For new "does it behave" coverage, we'll extend the @KSMLTest layer.

ksml/src/test/java/io/axual/ksml/operation/FilterOperationTest.java
pantelis17: Both stream tests now capture the FixedKeyProcessorSupplier and assert captor.getValue().get() is a FilterProcessor / FilterNotProcessor respectively, so they test genuinely different behavior and a wrong-processor bug fails. Also added FilterProcessorTest (mirror of FilterNotProcessorTest) to close the processor coverage gap.

ksml/src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1,2 @@
+# Auto-register test extensions declared under META-INF/services (see DefaultNotationIsolationExtension).
+junit.jupiter.extensions.autodetection.enabled=true
```
KasparMetsa: **Code in PR:**

```properties
# ksml/src/test/resources/junit-platform.properties
junit.jupiter.extensions.autodetection.enabled=true
```
```
# ksml/src/test/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
io.axual.ksml.execution.DefaultNotationIsolationExtension
```

This applies `DefaultNotationIsolationExtension` to **every** test in the `ksml` module (~700),
although only the ~38 `OperationTestSupport`-based tests register a mock default notation. It also
turns on a module-wide switch: any future extension declared under `META-INF/services` will silently
auto-register for all tests.

**Suggested change:** scope the extension to the tests that need it, without the global switch:

```java
// a small shared base or a custom @ExtendWith annotation used by the operation tests
@ExtendWith(DefaultNotationIsolationExtension.class)
abstract class OperationTestBase { }
```

**Why:** narrower blast radius and no surprise auto-registration for later contributors

ksml/src/test/java/io/axual/ksml/operation/OuterJoinWithStreamOperationTest.java
KasparMetsa: Thanks, that makes sense, and I agree on the join families - the table-join classes have several distinct scenarios each, so explicit classes read better than a 2-D parameterization. Let's keep those as they are.

For the single-method families (Convert* and the single-method Transform* classes), though, could you make that change in this PR rather than a follow-up? They are one test differing by a single verify(...) line, you already said they parameterize cleanly, and it is a small, contained edit - so it is cheaper to do now than to track and revisit in a separate PR.

On failure localization: JUnit reports the failing case by its argument/display name, so a named parameterized test still tells you exactly which operation failed. So we do not lose that here.

So: keep the join families as-is, and fold the Convert*/Transform* single-method ones into parameterized tests in this PR please. Thanks!

KasparMetsa (changes requested):

ksml/src/test/java/io/axual/ksml/execution/DefaultNotationIsolationExtension.java
```diff
@@ -0,0 +1,57 @@
...
+ * that registers a (mock) default notation cannot leak it into other test classes or make the suite
+ * order dependent.
+ *
+ * <p>Auto-registered for the whole module via {@code META-INF/services} and
+ * {@code junit-platform.properties}; it is a no-op for tests that never touch the default notation.
+ *
+ * <p>The notation library is a JVM-wide singleton, so this guards sequential execution only. Parallel
+ * test execution must stay off for this module.
+ */
+public class DefaultNotationIsolationExtension implements BeforeEachCallback, AfterEachCallback {
```
KasparMetsa, line 42: ## `DefaultNotationIsolationExtension` Javadoc still describes the removed auto-registration

The thread-8 fix deleted `junit-platform.properties` and the `META-INF/services` entry and switched to `@ExtendWith` on `OperationTestBase`. But the extension's own Javadoc was not updated and still describes the old mechanism.

**PR code (`DefaultNotationIsolationExtension.java`):**

```java
/**
 * ...
 * <p>Auto-registered for the whole module via {@code META-INF/services} and
 * {@code junit-platform.properties}; it is a no-op for tests that never touch the default notation.
 *
 * <p>The notation library is a JVM-wide singleton, so this guards sequential execution only. Parallel
 * test execution must stay off for this module.
 */
public class DefaultNotationIsolationExtension implements BeforeEachCallback, AfterEachCallback {
```

Neither `META-INF/services` nor `junit-platform.properties` exists anymore, and the extension is no longer auto-registered for the whole module.

**Suggested change:**

```java
/**
 * ...
 * <p>Applied only to the operation unit tests that register a mock default notation, by extending
 * {@code OperationTestBase} (which carries {@code @ExtendWith(DefaultNotationIsolationExtension.class)}).
 * It is a no-op for any test that never touches the default notation.
 *
 * <p>The notation library is a JVM-wide singleton, so this guards sequential execution only. Parallel
 * test execution must stay off for this module.
 */
```

**Reason:** the comment now says the opposite of what the code does. The next contributor reading it will look for a `META-INF/services` file that is not there, or assume every test is isolated when only `OperationTestBase` subclasses are. Wrong doc is worse than no doc.

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #657, opened 2026-07-09
fix Mockito java agent warnings

tonvanbart:
Change project pom files so that java agent is attached by -javaagent rather than auto-attaching.
 - ensure every module inherits Mockito
 - resolve Mockito coord to real jar path
 - add jar to argline as per docs
 - bring ksml-runner config in line with others


pom.xml
```diff
@@ -652,7 +666,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven.surefire.plugin.version}</version>
                     <configuration>
-                        <argLine>@{surefireArgLine} ${jdk24PlusJvmArgs}</argLine>
+                        <argLine>@{surefireArgLine} ${jdk24PlusJvmArgs} -javaagent:${org.mockito:mockito-core:jar}</argLine>
```
KasparMetsa, line 669: I tested this Claude suggestion locally and it worked well.

### 1. Use the Maven local repository path instead of `dependency:properties` to resolve the agent jar

The PR resolves the Mockito jar path using the `dependency:properties` goal. That goal only works if `mockito-core` is on the module's classpath, so the PR forces `mockito-core` onto every module by moving it from `<dependencyManagement>` (opt-in) to root `<dependencies>` (automatic for all). This silently adds Mockito to modules that deliberately had none, like `ksml-data`, and also attaches the Mockito ByteBuddy agent to every test JVM in the project - including modules that never use Mockito. The agent instruments classes at JVM startup, so modules with many small fast tests (such as `ksml-data` with 200+ tests) pay the agent startup cost on every run even though it does nothing for them.

A simpler approach that avoids the global injection and the unnecessary overhead: reference the jar directly via `${settings.localRepository}`, which is a built-in Maven variable pointing to `~/.m2/repository`. The path is stable - Maven always stores jars at `<localRepository>/<groupId path>/<artifactId>/<version>/<artifactId>-<version>.jar`.

---

#### File 1: `pom.xml` (root)

**Change A - keep `mockito-core` in `<dependencyManagement>`, do not move it**

The PR removes it from `<dependencyManagement>` (around line 547) and adds it to root `<dependencies>`. Do not make that change. Leave the block exactly as it is:

```xml
<!-- pom.xml - leave this in <dependencyManagement>, do not touch it -->
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <version>${mockito-core.version}</version>
    <scope>test</scope>
</dependency>
```

**Reason:** `<dependencyManagement>` is opt-in - a module only gets the dependency if it explicitly asks for it. This preserves the intentional choice of modules like `ksml-data` to not use Mockito. The version is still managed centrally, nothing changes for modules that do declare it.

**Change B - do not add `mockito-core` to root `<dependencies>`**

The PR adds a new root `<dependencies>` block. Do not add it:

```xml
<!-- PR adds this - do NOT add it -->
<dependencies>
    <dependency>
        <groupId>org.mockito</groupId>
        <artifactId>mockito-core</artifactId>
        <version>${mockito-core.version}</version>
        <scope>test</scope>
    </dependency>
</dependencies>
```

**Reason:** Root `<dependencies>` is automatically inherited by every child module. This silently adds Mockito to modules that have no tests, or that deliberately chose not to use it. A developer in any module can now write Mockito tests without declaring the dependency, making the dependency invisible in that module's `pom.xml`. If this global injection is later removed, those tests break with no clear indication of why.

**Change C - do not add the `mockito-agent-property` execution to `maven-dependency-plugin`**

The PR adds an extra execution to `maven-dependency-plugin`:

```xml
<!-- PR adds this inside maven-dependency-plugin - do NOT add it -->
<execution>
    <id>mockito-agent-property</id>
    <goals>
        <goal>properties</goal>
    </goals>
</execution>
```

**Reason:** This execution exists only to set the `${org.mockito:mockito-core:jar}` property used in the PR's argLine approach. With the `${settings.localRepository}` approach in Change E, the property is not needed, so this execution serves no purpose.

**Change D - do not add `maven-dependency-plugin` to root `<build><plugins>`**

The PR adds this to the root `<build><plugins>` section. Do not add it:

```xml
<!-- PR adds this to <build><plugins> - do NOT add it -->
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-dependency-plugin</artifactId>
</plugin>
```

**Reason:** This entry activates the `mockito-agent-property` execution from Change C for every module. Since Change C is not added, this entry has nothing to activate and should not be added either.

**Change E - in `<pluginManagement>`, update the surefire `argLine`** (line 655 in the current file)

**PR version:**
```xml
<argLine>@{surefireArgLine} ${jdk24PlusJvmArgs} -javaagent:${org.mockito:mockito-core:jar}</argLine>
```

**Correct version:**
```xml
<argLine>@{surefireArgLine} ${jdk24PlusJvmArgs} -javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito-core.version}/mockito-core-${mockito-core.version}.jar</argLine>
```

**Reason:** `${org.mockito:mockito-core:jar}` requires the `dependency:properties` goal to run first, which requires `mockito-core` to be on the classpath, which requires the global injection (changes A-D above). `${settings.localRepository}` resolves directly without any of that machinery. Tested: zero Mockito warnings, `BUILD SUCCESS`.

---

#### File 2: `ksml-runner/pom.xml`

**Remove the module-level surefire override** (lines 135-141 in the current file).

`ksml-runner` has its own surefire block that duplicates the root argLine but without the Mockito agent. Removing it lets the module inherit the corrected root argLine.

**PR code (correct - keep this change):**
```xml
<!-- remove this entire block from ksml-runner/pom.xml -->
-<plugin>
-    <groupId>org.apache.maven.plugins</groupId>
-    <artifactId>maven-surefire-plugin</artifactId>
-    <configuration>
-        <argLine>@{surefireArgLine} ${jdk24PlusJvmArgs}</argLine>
-    </configuration>
-</plugin>
```

The PR already removes this, which is correct. Keep that removal.

**Reason:** `ksml-runner` previously had its own surefire block that duplicated the root argLine without the Mockito agent. This override was the reason the agent did not apply to `ksml-runner` even after changing the root pom. Removing it lets `ksml-runner` inherit the correct argLine from the root, including the agent.

---

#### All other PR changes - keep them

The PR also removes explicit `mockito-core` declarations from several individual module `pom.xml` files (`ksml-data-avro-apicurio`, `ksml-data-avro-confluent`, `ksml-data-jsonschema-apicurio`, `ksml-data-protobuf-apicurio`, `ksml-runner`, `ksml`). Those removals are correct - those modules get `mockito-core` transitively through `mockito-junit-jupiter` and the explicit declarations were redundant.

KasparMetsa (changes requested):

pom.xml
tonvanbart: This will (probably) work, and I like that it removes the mockito dependency from modules that don't actually need it. However, it comes with it's own risks:

- If `mockito,core.version` happens to be any `-SNAPSHOT`, the actual file name in the repo is timestamped, and this solution will not resolve to any file
- the effective local repo can be overridden via local settings, `-Dmaven.repo.local`, wrapper scripts etc which could point to stale versions
- on Windows, concatenating the path like this can produce mixed separators or a path with spaces in it, which will fail

The solution using the dependency plugin `properties` goal sidesteps these.

I agree that this solution is a compromise in that it adds mockito-core to modules that don't need it, and that it will make the build slightly slower (but carries no runtime penalty). But I suggest to keep this rather than switching to a more brittle solution. WDYT?

(For example https://youtrack.jetbrains.com/issue/IDEA-388431/Windows-test-run-fails-when-the-Mockito-javaagent-path-from-settings.localRepository-contains-spaces documents the "spaces in path" issue)

KasparMetsa: I agree, let's keep the `dependency:properties` way.

Btw, this is the way the Mockito docs tell us to set up the agent on Java 21 and newer: https://github.com/mockito/mockito/blob/main/mockito-core/src/main/java/org/mockito/Mockito.java

Section 0.3 of the class Javadoc says:
> To add Mockito as an agent to Maven's surefire plugin, the following configuration is needed"

and it then shows the `maven-dependency-plugin` properties goal together with this exact line: `<argLine>@{argLine} -javaagent:${org.mockito:mockito-core:jar}</argLine>`. 
The properties goal is what fills in that `${org.mockito:mockito-core:jar}` path. So we are just following the official guide.

The properties goal also finds the real jar file and makes sure Maven has downloaded it first. My ${settings.localRepository} idea builds the file path by hand and does not download anything. 

Right now Mockito is added to all 16 modules, but only 5 actually use it. The current PR is the simple option and it follows the official pattern, so I am happy to go with it.

KasparMetsa (approved):

PR #660, opened 2026-07-14
Add unit tests for Avro and Protobuf schema mapper classes

pantelis17:
# Add unit tests for the KSML data-type modules (AVRO and PROTOBUF)

## What this PR does

This PR adds unit tests for the schema mappers in the data-type modules. These mappers convert between an external schema format (Avro or Protobuf) and the internal KSML `DataSchema` model, in both directions. Before this PR these classes had no direct unit tests.

No production code changes — this is tests only, plus the small build wiring needed to run them.

## New tests

| Module | Test | What it checks |
|---|---|---|
| `ksml-data-avro` | `AvroSchemaMapperTest` | Avro ↔ KSML schema conversion: primitives, collections, unions, logical types, optional fields, enums, records, field order, and default values. Also regression tests for the record/array default-value handling (the `NULL_VALUE` sentinel and "no `Unknown datum class`" bug). |
| `ksml-data-protobuf` | `ProtobufFileElementSchemaMapperTest` | Protobuf `ProtoFileElement` ↔ KSML schema: primitive type groups, repeated fields, `oneof`, enums, nested messages, proto2/proto3 optionality, defaults, error cases, and nested-type de-duplication. |
| `ksml-data-protobuf-apicurio` | `ApicurioProtobufFileElementDescriptorMapperTest` | Apicurio descriptor mapper: `ProtoFileElement` ↔ protobuf `FileDescriptor`. |
| `ksml-data-protobuf-confluent` | `ConfluentProtobufFileElementDescriptorMapperTest` | Confluent descriptor mapper: same conversions, verified for the Confluent implementation. |

Most tests do a round-trip (`external → KSML → external`) and assert the result is stable, plus they check the actual field types, names, labels and defaults — not just that an object comes back.

## Shared test base (no duplication between Apicurio and Confluent)

The Apicurio and Confluent descriptor mappers do almost the same job, so their tests would be almost identical. To avoid copy-paste, the shared test body lives once in an abstract base class:

- `AbstractProtobufFileElementDescriptorMapperTest` in the `ksml-data-protobuf` **test-jar** holds all the shared test cases and the helper builders.
- Each module has a small subclass that only provides its own mapper and the one assertion that genuinely differs (Confluent exposes `oneof` fields differently because it goes through `DynamicSchema`).

## Build changes

- `ksml-data-protobuf/pom.xml`: build a `test-jar` so the two downstream modules can reuse the shared base test.
- `ksml-data-protobuf-apicurio/pom.xml` and `ksml-data-protobuf-confluent/pom.xml`: depend on that `test-jar` (test scope).
- `ksml-data-protobuf-confluent/pom.xml`: add the `junit-jupiter` test dependency (this module had no tests before, so it was not present).

## Notes

- All tests use AssertJ and have `@DisplayName` for readable reports.
- All four modules build and pass green.


ksml-data-protobuf-confluent/src/test/java/io/axual/ksml/data/notation/protobuf/ConfluentProtobufFileElementDescriptorMapperTest.java
```diff
@@ -0,0 +1,53 @@
...
+    protected ProtoFileElement toFileElement(Descriptors.Descriptor descriptor) {
+        return mapper.toFileElement(descriptor);
+    }
+
+    @Override
+    protected void assertOneOfBranches(Descriptors.Descriptor msgDescriptor) {
+        // Confluent builds the descriptor via DynamicSchema, which links oneof fields to the group
+        // differently from Apicurio, so verify field accessibility by name rather than via the
+        // oneof's field list.
+        assertThat(msgDescriptor.findFieldByName("text")).isNotNull();
```
KasparMetsa: **Code in PR** (`ConfluentProtobufFileElementDescriptorMapperTest.assertOneOfBranches`):

```java
// Confluent builds the descriptor via DynamicSchema... verify field accessibility by name
assertThat(msgDescriptor.findFieldByName("text")).isNotNull();
assertThat(msgDescriptor.findFieldByName("text").getType()).isEqualTo(Descriptors.FieldDescriptor.Type.STRING);
assertThat(msgDescriptor.findFieldByName("count")).isNotNull();
assertThat(msgDescriptor.findFieldByName("count").getType()).isEqualTo(Descriptors.FieldDescriptor.Type.INT32);
```

The Apicurio version asserts the fields are actually grouped in a `oneof`
(`getOneofs().hasSize(1)`, name `payload`, 2 fields). The Confluent version only checks the two fields
exist with the right types - it does **not** verify they are in a `oneof` at all. If the Confluent
mapper flattened the `oneof` into two plain fields, this test would still pass, even though the whole
point of the case is `oneof` handling.

**Suggested change:** verify the grouping through the field's containing oneof, which protobuf-java
exposes regardless of how the descriptor was built:

```java
assertThat(msgDescriptor.findFieldByName("text").getContainingOneof()).isNotNull();
assertThat(msgDescriptor.findFieldByName("text").getContainingOneof().getName()).isEqualTo("payload");
assertThat(msgDescriptor.findFieldByName("count").getContainingOneof().getName()).isEqualTo("payload");
```

**Why:** it keeps the two vendor tests equivalent in strength, and it actually tests the `oneof`
behavior the case is named for. If `getContainingOneof()` genuinely does not work on the Confluent
path, add a comment stating the grouping cannot be verified there, so the weaker check is a documented
choice.

ksml-data-protobuf-apicurio/pom.xml
```diff
@@ -23,6 +23,13 @@
             <groupId>io.axual.ksml</groupId>
             <artifactId>ksml-data-protobuf</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.axual.ksml</groupId>
+            <artifactId>ksml-data-protobuf</artifactId>
```
KasparMetsa, line 33: ## Apicurio test module has no direct JUnit dependency (inconsistent with Confluent, and brittle)

The two downstream modules are wired differently for the same job.

Confluent pom adds an explicit JUnit dependency:

```xml
<!-- ksml-data-protobuf-confluent/pom.xml -->
<dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter</artifactId>
    <scope>test</scope>
</dependency>
```

Apicurio pom does NOT. It only adds the shared test-jar:

```xml
<!-- ksml-data-protobuf-apicurio/pom.xml -->
<dependency>
    <groupId>io.axual.ksml</groupId>
    <artifactId>ksml-data-protobuf</artifactId>
    <version>${project.version}</version>
    <type>test-jar</type>
    <scope>test</scope>
</dependency>
```

The new Apicurio test still compiles and runs today only because `mockito-junit-jupiter` (already in that pom) drags in the JUnit API by transitive path. The test class itself never uses Mockito.

Suggested change: add the same explicit `junit-jupiter` test dependency to `ksml-data-protobuf-apicurio/pom.xml` that Confluent has.

Reason: a test class should declare the framework it uses. Right now the Apicurio descriptor-mapper test depends on Mockito staying in the pom. If someone later removes Mockito from that module (the new test does not need it), the test stops compiling for a reason that is hard to see. Matching Confluent also keeps the two sibling modules consistent, which is the pattern everywhere else in this repo.

ksml-data-protobuf/src/test/java/io/axual/ksml/data/notation/protobuf/ProtobufFileElementSchemaMapperTest.java
```diff
@@ -0,0 +1,547 @@
...
+                .findFirst().orElseThrow();
+        final var greetingField = msgElement.getFields().stream()
+                .filter(f -> "greeting".equals(f.getName())).findFirst().orElseThrow();
+        assertThat(greetingField.getDefaultValue()).isEqualTo("hello");
+    }
+
+    // ===== fromDataSchema: MapSchema is silently dropped =====
+
+    @Test
+    @DisplayName("MapSchema field is dropped from the output (proto does not support Avro-style maps via this path)")
```
KasparMetsa: ## Test pins a silent field-drop as "intended", with a comment that is not accurate

```java
// ProtobufFileElementSchemaMapperTest.java
@Test
@DisplayName("MapSchema field is dropped from the output (proto does not support Avro-style maps via this path)")
void fromDataSchema_mapSchemaField_silentlyDropped() {
    // Intended behavior: convertDataSchemaToProtoType returns null for a MapSchema, and
    // convertStructSchemaToMessageElement skips fields with a null proto type. This test pins
    // that contract so a future change to map handling is a conscious decision, not a surprise.
    ...
    assertThat(msgElement.getFields()).hasSize(1);
    assertThat(msgElement.getFields().getFirst().getName()).isEqualTo("kept");
}
```

The production code does drop the field (`ProtobufFileElementSchemaMapper` returns `null` for `MapSchema`, and the field with a null type is skipped), so the test is factually correct about current behavior.

Two problems:

- The comment "proto does not support Avro-style maps via this path" is misleading. Proto3 does support maps (`map<key, value>`). The real story is that this mapper does not implement map output yet, so the field is dropped without any error or warning. That is closer to a known gap than to a designed feature.
- A green test named `...silentlyDropped` freezes a data-loss behavior in place. Later, anyone who fixes map support has to also change a passing test, which reads as if the drop was wanted.

Suggested change: keep the test (pinning current behavior is fine), but reword the `@DisplayName` and comment to say this is a current limitation, not a feature, and link a tracking issue. For example: "MapSchema field is currently not written to proto and is dropped (known limitation, see ISSUE-xxx)". If a ticket does not exist, create one.

Reason: the test is useful as a guard, but the wording should not make a missing feature look intentional. Clear wording stops the next reader from treating silent data loss as correct.

ksml-data-protobuf/pom.xml
```diff
@@ -60,6 +60,18 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-failsafe-plugin</artifactId>
             </plugin>
+            <!-- Publish a test-jar so downstream protobuf modules can reuse the shared descriptor-mapper test base -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
```
KasparMetsa, line 74: ## New `test-jar` + abstract-base pattern has no precedent in this repo; document the coupling

This PR introduces two things that did not exist in the codebase before:

- The first `test-jar` published by any module (`ksml-data-protobuf/pom.xml`).
- The first `Abstract...Test` shared test base (`AbstractProtobufFileElementDescriptorMapperTest`).

```xml
<!-- ksml-data-protobuf/pom.xml -->
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-jar-plugin</artifactId>
    <executions>
        <execution>
            <goals>
                <goal>test-jar</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

The approach is reasonable and avoids copy-paste between Apicurio and Confluent, and it works in CI (`mvn clean verify`) and with `-am`. But it adds build coupling: the two downstream test modules now cannot build their tests unless `ksml-data-protobuf` is available as a `test-jar`. Running a single module without `-am` and without a prior `mvn install` (for example `mvn test -pl ksml-data-protobuf-confluent`) will fail to resolve the test-jar.

Suggested change: no code change required, but add a short note (in the module README or a comment near the `test-jar` plugin block) explaining that these two modules share a test base via the test-jar, and that building them alone needs `-am` or a prior install. Consider whether a smaller shared helper (a non-test utility, or duplicating ~40 lines) would avoid the cross-module test-jar dependency, since the shared subclasses are tiny.

Reason: this is a new build convention for the project. A one-line note saves the next developer from a confusing "cannot resolve test-jar" error and records why the pattern was chosen.

KasparMetsa (changes requested):

ksml-data-protobuf-confluent/src/test/java/io/axual/ksml/data/notation/protobuf/ConfluentProtobufFileElementDescriptorMapperTest.java
pantelis17: Tried getContainingOneof() as suggested, but it returns null on the Confluent path: DynamicSchema doesn't retain the oneof grouping on the built descriptor (both getOneofs() and getContainingOneof() are empty/null there, unlike Apicurio). So per your fallback, I've kept the field-existence + type check and documented the limitation explicitly in the test.

ksml-data-protobuf/src/test/java/io/axual/ksml/data/notation/protobuf/ProtobufFileElementSchemaMapperTest.java
pantelis17: Done and linked with this https://github.com/Axual/ksml/issues/661

KasparMetsa (approved):

PR #663, opened 2026-07-15
Speedup unit and integration tests

pantelis17:
# Speed up the unit and integration test suites

## What this PR does

`mvn clean verify` was slow (around 9 minutes). This PR makes it faster in two ways, without
changing what any test actually checks.

### 1. Build the modules in parallel

Added a `.mvn/maven.config` file with `-T1C`. This tells Maven to use one build thread per CPU
core, so the many Maven modules build and run their unit tests in parallel instead of one by one.

Because it lives in `.mvn/maven.config`, it works everywhere: local `mvn` commands and the CI
pipeline both pick it up automatically. You can still force a single thread with `mvn -T1 ...`.

### 2. Share the Kafka broker and schema registry between integration tests

Before, every integration test (IT) started its **own** Kafka container and its **own** Apicurio
schema registry container. Starting these containers is the slowest part of the IT suite, and we
were paying that cost around 10 times.

Now all ITs share **one** Kafka broker and **one** Apicurio schema registry through a new helper
class, `SharedKsmlInfra`. The containers start once (lazily, on first use) and are reused by every
IT. Testcontainers' Ryuk process cleans them up when the JVM exits.

## Why the topic renames

Because the broker is now shared, two ITs can no longer use the same topic name, or they would
read each other's data. The Confluent AVRO test used the same topic names as the Apicurio AVRO
test, so its topics were renamed to be unique:

- `sensor_data_avro` -> `sensor_data_avro_confluent`
- `sensor_data_transformed` -> `sensor_data_transformed_confluent`

To keep the test data identical, a copy of the AVRO example files was added under
`avro-confluent/`. These files are the same as the original AVRO example; only the topic names
differ. The test scenario (produce sensor data -> uppercase the name -> check the output) is
unchanged.

## Other changes

- **`KSMLContainer.createKafkaTopics`** now only creates topics that do not exist yet, since the
  broker is shared. It also tolerates a topic being created at the same time instead of failing.
- **`AvroSchemaEvolutionBackwardCompatibilityIT`** and the other ITs dropped their private Kafka
  and registry fields and now use `SharedKsmlInfra`.
- **README.md** and **CONTRIBUTING.md** now explain the `-T1C` default and how to opt out.

## Safety notes

- Every IT keeps the exact same inputs, pipeline logic, assertions, and wait conditions. The only
  functional change is the Confluent topic names described above.
- All ITs use unique topic names, so they do not collide on the shared broker or on schema-registry
  subjects.
- Within one JVM the tests still run sequentially (the default), so sharing the containers is safe.

## How it was verified

- `mvn -pl ksml-integration-tests clean test-compile` builds cleanly.
- The full IT suite runs green with one Kafka container and one registry container instead of many.


pantelis17:
Now locally is: `mvn clean verify --no-transfer-progress  306.90s user 22.54s system 157% cpu 3:29.48 total` vs `mvn clean verify --no-transfer-progress  291.11s user 20.10s system 94% cpu 5:30.69 total` before.

Also, in CI, now executed in `6m 29s` while before in `7m 43s`. For more speed up we need to run the pipeline with stronger runner e.g. 4, 8 or more CPU threads

ksml-integration-tests/src/test/java/io/axual/ksml/integration/ConfluentAvroSchemaRegistryIT.java
```diff
@@ -50,28 +48,14 @@
...
+            .withKsmlFiles("/docs-examples/beginner-tutorial/different-data-formats/avro-confluent",
                           "producer-avro.yaml", "processor-avro-transform.yaml", "SensorData.avsc")
-            .withKafka(kafka)
-            .withConfluentAvroRegistry(schemaRegistry)
-            .withTopics("sensor_data_avro", "sensor_data_avro_transformed")
-            .dependsOn(kafka, schemaRegistry);
-
+            .withKafka(SharedKsmlInfra.kafka())
+            .withConfluentAvroRegistry(SharedKsmlInfra.schemaRegistry())
+            .withTopics("sensor_data_avro_confluent", "sensor_data_transformed_confluent")
```
KasparMetsa, line 57: ## The documented Confluent AVRO example is no longer tested

This is the most important one. It is not a build break, but it quietly removes a
guarantee the repository was built to give.

The docs pull their example files straight from the integration-test resources with
`include`. In `docs/tutorials/beginner/data-formats.md`:

```
line  83: include ".../different-data-formats/avro/confluent_avro/ksml-runner.yaml"
line 159: include ".../different-data-formats/avro/processor-avro-transform.yaml"
```

Before this PR, `ConfluentAvroSchemaRegistryIT` ran exactly those two files, so CI
proved the documented Confluent example actually works.

This PR points the test at a fresh copy instead:

```java
// ConfluentAvroSchemaRegistryIT.java (in the PR)
static final KSMLContainer ksml = new KSMLContainer()
        .withKsmlFiles("/docs-examples/beginner-tutorial/different-data-formats/avro-confluent",
                      "producer-avro.yaml", "processor-avro-transform.yaml", "SensorData.avsc")
        ...
        .withTopics("sensor_data_avro_confluent", "sensor_data_transformed_confluent")
```

The PR does not touch `data-formats.md`, so the docs still include the old
`avro/confluent_avro/ksml-runner.yaml` and `avro/processor-avro-transform.yaml`.
After the merge, no integration test runs those two files. If someone edits the
documented files and breaks them, CI stays green. The test now proves a near-copy,
not the thing the reader is shown.

Suggested change:

- Point the two `include` lines in `data-formats.md` at the new `avro-confluent`
  copies, so the documented files are again the tested files. This keeps the
  doc/test link intact with the least effort.

Why it matters: the whole point of keeping the examples under
`ksml-integration-tests/src/test/resources/docs-examples` and pulling them into the
docs with `include` is that the docs are runnable and CI-verified. Breaking that link
for one example invites silent drift between what is documented and what works.

ksml-integration-tests/src/test/java/io/axual/ksml/integration/testutil/SharedKsmlInfra.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return kafka;
+    }
+
+    /** The shared Apicurio schema registry (legacy-id mode), started on first access. */
+    public static synchronized ApicurioSchemaRegistryContainer schemaRegistry() {
+        if (schemaRegistry == null) {
+            schemaRegistry = new ApicurioSchemaRegistryContainer()
+                    .withNetwork(NETWORK)
+                    .withLegacyIdMode();
```
KasparMetsa, line 79: ## Silent change to how two tests talk to the schema registry

The PR description says the only functional change is the Confluent topic names. That
is not quite true.

Before the PR the registry containers were not all the same:

```java
// old ConfluentAvroSchemaRegistryIT and old JsonSchemaRegistryIT
static final ApicurioSchemaRegistryContainer schemaRegistry =
        new ApicurioSchemaRegistryContainer().withNetwork(network); // no legacy id mode
```

Now every registry-based test shares one registry that always has legacy id mode on:

```java
// SharedKsmlInfra.java (in the PR)
schemaRegistry = new ApicurioSchemaRegistryContainer()
        .withNetwork(NETWORK)
        .withLegacyIdMode(); // sets APICURIO_CCOMPAT_LEGACY_ID_MODE_ENABLED=true
```

So `ConfluentAvroSchemaRegistryIT` and `JsonSchemaRegistryIT` now run against a
registry configured differently than before. Legacy id mode changes the id the
Confluent-compatible endpoint writes into the message header, which is exactly the
part the AVRO and JSON serdes depend on. The suite is reported green, so it seems to
work, but this is a real environment change that is not called out.

Suggested change:

- Add one line to the PR description under "the only functional change" that says the
  JSON and Confluent tests now run against a legacy-id-mode registry, and confirm this
  is intended.
- Optionally note in the `SharedKsmlInfra` Javadoc that legacy id mode is required by
  the Apicurio-native tests and is safe for the Confluent-compatible ones, so a later
  reader does not "clean it up".

Why it matters: bundling a quiet behavior change into a "speedup" PR makes it hard to
trace later if a serde test starts failing. It should be visible in the description.

.mvn/maven.config
```diff
@@ -0,0 +1 @@
+-T1C
```
KasparMetsa: ## `-T1C` is applied to every Maven command, not just the test build

```
# .mvn/maven.config (in the PR)
-T1C
```

Because it lives in `.mvn/maven.config`, it affects every invocation for every
contributor and every pipeline: `mvn clean package`, the release/deploy build, and
`mvn sonar:sonar`, not only `mvn clean verify`. The README and CONTRIBUTING notes are
good, but the blast radius is wider than "the test suite".

Concerns:

- The `ksml` module tests run Python through GraalVM. Building and unit-testing many
  modules at once raises peak memory. CI runs on `ubuntu-latest`, so this needs a
  quick check that the runner has enough memory, or a job could start failing in a way
  that is hard to reproduce locally.
- Parallel module builds interleave log output, which makes a failing build harder to
  read.
- Turning it on for the release/deploy build changes how artifacts are produced. Some
  plugins are not thread-safe; this should be confirmed against the actual deploy job.

Suggested change:

- Keep the speedup, but verify the CI run with `-T1C` on the real runner (memory and a
  green integration run), and confirm the deploy/sonar jobs are happy with it.
- If only the test build needs it, consider passing `-T1C` in the CI test step instead
  of the always-on `.mvn/maven.config`, so local single-threaded debugging stays the
  default and the release build is untouched.

Why it matters: an always-on reactor setting is convenient but easy to forget; it
should be a deliberate choice for the release and sonar paths too, not a side effect.

ksml-integration-tests/src/test/java/io/axual/ksml/integration/testutil/KSMLContainer.java
```diff
@@ -312,13 +314,28 @@ private void createKafkaTopics() {
             props.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, kafka.getBootstrapServers());
 
             try (AdminClient adminClient = AdminClient.create(props)) {
-                List<NewTopic> topics = new ArrayList<>();
+                // The broker is shared across IT classes, so only create topics that do not exist yet
+                // (createTopics fails on an already-existing topic).
+                final var existingTopics = adminClient.listTopics().names().get();
```
KasparMetsa, line 319: ## `createKafkaTopics` silently ignores partition count on existing topics

```java
// KSMLContainer.createKafkaTopics (in the PR)
final var existingTopics = adminClient.listTopics().names().get();
final var topics = new ArrayList<NewTopic>();
for (String topicName : topicsToCreate) {
    if (!existingTopics.contains(topicName)) {
        topics.add(new NewTopic(topicName, topicPartitionCount, (short) 1));
    }
}
```

The "create only if missing" logic is correct for a shared broker. The gap: if a topic
already exists with a different partition count than requested, it is skipped without a
word. `RepartitionIT` calls `withPartitions(4)` and depends on 4 partitions. Today all
topic names are unique per test, so this cannot bite, but it is a quiet trap for the
next person who reuses a topic name.

Suggested change:

```java
if (!existingTopics.contains(topicName)) {
    topics.add(new NewTopic(topicName, topicPartitionCount, (short) 1));
} else if (topicPartitionCount > 1) {
    // A shared topic kept from an earlier test may have fewer partitions than we need.
    log.warn("Topic {} already exists; not changing its partition count (wanted {})",
            topicName, topicPartitionCount);
}
```

Why it matters: a one-line warning turns a silent, confusing failure into an obvious
one if topic names ever overlap.

KasparMetsa (changes requested):

KasparMetsa (approved):

PR #667, opened 2026-07-20
Upgrade Jackson to v3 & other major dependency upgrades

pantelis17:
# Major dependency upgrades (KSML 2.0.0)

This PR moves KSML to the newest versions of its main libraries. Most are big version jumps that change the library API, so a lot of code had to change too. This is the first step towards KSML 2.0.0. Most of the diff is repetitive: new package names, new `NOTICE.txt` files, new version numbers.

## Versions

| Library | Old | New | What it means |
|---|---|---|---|
| Jackson | 2.22.0 | 3.2.0 | Packages moved to `tools.jackson`. Every file that reads or writes JSON or YAML needed new imports |
| Apicurio Registry | 2.6.13 | 3.3.0 | REST path changed from `/apis/registry/v2` to `/v3`. Some serde libraries got new names |
| Protobuf | 3.25.9 | 4.35.1 | New major version of the protobuf runtime |
| Wire | 5.2.1 | 6.4.1 | The `ProtoFileElement` constructor has one more parameter |
| victools | 4.38.0 | 5.0.0 | The version that works with Jackson 3 |
| Confluent serdes | 8.x | 8.3.0 | Matches the new Apicurio version |
| logback / junit / others | | newer | Small and safe updates |

Apache Kafka is **not** upgraded. It stays on `4.3.1`, the same version as `main`.

## What users must change

The new [migration guide](https://github.com/Axual/ksml/blob/major-dependency-upgrades/docs/migration-to-2.0.0.md) shows each setting before and after.

- Apicurio registry URLs must use `/apis/registry/v3`.
- The basic-auth keys have new names. `apicurio.auth.*` becomes `apicurio.registry.auth.*`.
- `apicurio.registry.auto-register.if-exists: RETURN` no longer exists. Use `FIND_OR_CREATE_VERSION`.
- `apicurio.registry.as-confluent` no longer exists. `Legacy4ByteIdHandler` is now `Default4ByteIdHandler`. If KSML finds one of these old settings, it stops at startup and tells you. Before, the setting was dropped without any message.
- KSML now stops at startup if `ksml-runner.yaml` has a key it does not know, or the same key twice. The same is true for a definition file with the same key twice, or a file KSML cannot read. Before, these were ignored.

The bytes that Apicurio writes to Kafka do not change. You do not need to reprocess any topic.

## Jackson 2 is still in the build

All KSML code now uses Jackson 3. But the build still contains Jackson 2, because Apicurio brings it in. Apicurio is built on Quarkus, and Quarkus still lists 72 Jackson 2 libraries and no Jackson 3 ones. The runner goes from 207 to 244 third-party libraries.

The root `pom.xml` loads both Jackson BOMs (the files that fix library versions). Jackson 3 comes first, so it sets the version of `jackson-annotations`. Without this, Maven picks Jackson 2 versions that do not match each other.

## Things worth a closer look

- **New module `ksml-data-apicurio`.** A new published library. It holds `ApicurioConfigChecks`, the startup checks for old Apicurio v2 settings. All three Apicurio notations share it, instead of each keeping a copy.
- **Bug fix.** The remote schema lookup sent an empty group and version. The Apicurio v3 client does not accept that, so any stream with `valueType: avro` and no schema name failed at startup. It now sends the group from the config and asks for the latest version.
- **Bug fix.** In Jackson 3, `asString()` throws an error for an object or a list. Jackson 2 returned empty text. Three parsers now check the kind of value first, so a wrong nested block gives a normal KSML parse error again.
- **`@Jacksonized` removed.** It creates Jackson 2 annotations, which Jackson 3 ignores, so it already did nothing. Note: the next `@Builder` config class that is not a record will fail at runtime, and `@Jacksonized` will not help.
- **New JSON Schema mapper.** Apicurio 3 still uses Jackson 2 inside. So the JSON Schema notation has its own `ApicurioJsonSchemaDataObjectMapper`. It passes plain Java maps and lists, because a `JsonNode` from one Jackson version does not work with the other.
- **Less duplicate code.** Both protobuf modules had the same helper class. There is now one shared `DescriptorToFileElementConverter`.
- **`HeaderFilterSerde` deleted.** Only the Apicurio protobuf serde used it, and it filtered no headers at all. So no real users were left. **It was public API in `ksml-data`.** Tell us if another Axual project uses it, and we can bring it back.

## Tests

2603 tests pass, 0 failures. This includes all ten integration test classes, which run against real Kafka and Apicurio 3.3.0 containers. The containers now start once for all tests instead of once per test, which is faster.

New tests cover the stricter config rules, the rejected Apicurio v2 settings, the schema lookup, and the three parsers. The PR also adds a migration guide, and updates the release notes, the configuration reference and the `DEVELOPER_GUIDE.md` files.

## Why so many test files changed, and why that is safe

71 test files changed. That looks like a lot, so here is what changed and why no feature is at risk.

There are three kinds of change:

1. Renames only. For example `asText` became `asString`.
2. The same test, written again for the new API. The Kafka Streams error handler moved from `CONTINUE`, `FAIL` and `RETRY` to `Response.resume()`, `Response.fail()` and `Response.retry()`. Every case is still tested.
3. Tests that changed because the class they test changed. The fake registry client went from 75 methods to 13, because the Apicurio v3 interface is much smaller.

The most important tests got stronger. The three Apicurio serde tests now check `ID_HANDLER` and `FIND_LATEST_ARTIFACT`. The old tests never checked these, and they are what keeps the bytes on Kafka the same.

The ten integration test classes also still run the features from start to end against real containers, and they pass.

## Later work

None of these block this PR.

- Remove Jackson 2 from the build, after Quarkus and then Apicurio move to Jackson 3.
- Replace the `ConsumerGroupMetadata` constructor that Kafka marked for removal, when Kafka offers a new one.
- Move `ResolvingAdmin` off the old `listConsumerGroups` API. This changes behaviour for anyone who uses `ksml-kafka-clients` as a library.
- Make Apicurio nested type resolution work the same way in all three notations.
- Measure the JSON Schema read path. Today each record goes through a JSON text step, to move between the two Jackson versions.


pantelis17:
More details for the follow up tickets:
[PR-667-followup-tickets.md](https://github.com/user-attachments/files/30421157/PR-667-followup-tickets.md)

tonvanbart (approved):

jeroenvandisseldorp (approved):

PR #669, opened 2026-07-21
Refactor parsers

tonvanbart:
- Claude assisted refactoring to fix Sonar warnings related to deep inheritance registries. See `openspec/changes/simplify-baseparser-composition` for details.
 - Added some Javadoc
 - added developerdocs/parser-architecture.md 
 - added developerdocs/bugs-and-dead-code-canditates.md


ksml/src/test/java/io/axual/ksml/parser/StructsParserLazyTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.axual.ksml.data.schema.StructSchema;
+import io.axual.ksml.generator.YAMLObjectMapper;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
pantelis17: As far as i know, we are using AssertJ in our unit tests so please update the tests to use it instead of JUnit

tonvanbart: I can change this one, but a quick check shows that none of the other tests seem to use AssertJ. FYI

developerdocs/parser-architecture.md
```diff
@@ -0,0 +1,338 @@
...
+| `FixedSchema`  | A fixed-length byte sequence                                        |
+| `TupleSchema`  | A fixed-arity heterogeneous sequence                                 |
+
+A `DataSchema` value carries no parsing behavior — it's pure data, safely inspectable and
+serializable. That's what makes it the right thing for `schema()`/`schemas()` to return: the
+description can be extracted, combined, and converted **without ever calling `parse()`**.
+
+### Where `schema()` gets its information from
+
+The field-builder methods that build a `StructsParser<T>` — `stringField`, `booleanField`,
```
KasparMetsa, line 121: ### Current problem

This section describes the field-builder methods (`stringField`, `structsParser`, `optional`, ...)
in two ways that cannot both be true. Lines 121-132 say they are `protected` instance methods on
`DefinitionParser<T>`, inherited by every subclass. Lines 134-141 then call the move to
`FieldParsers` a "proposed improvement" in a "companion change". This PR already made that move:
`DefinitionParser.java` is 41 lines and has none of them. Section 7 has the same split. The example
at lines 252-262 uses short calls, but `definition/parser/TopologyDefinitionParser.java:50` reads
`FieldParsers.optional(FieldParsers.mapField(...))`.

### Suggested fix

Describe only the design that this PR ships. Rewrite the text under "Where `schema()` gets its
information from" and the note at lines 289-294, change the example to `FieldParsers.xxx(...)`, and
delete the "Proposed improvement" block together with every note about which branch the reader is
on. Those notes will be wrong on every branch soon, and the reason for the change is already in the
openspec proposal. This also makes the style match `bugs-and-dead-code-candidates.md`, which states
such facts plainly.

ksml/src/main/java/io/axual/ksml/definition/PythonSource.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+// The three kinds of Python source text a FunctionDefinition can carry: global setup code,
+// the function body, and a return expression. Bundled into one type so FunctionDefinition's
+// factory methods and constructor take one parameter for "the code" instead of three,
+// keeping their parameter counts under the java:S107 threshold.
+
+import java.util.Arrays;
+
+public record PythonSource(String[] globalCode, String[] code, String[] expression) {
```
KasparMetsa: ### Current problem

The record holds three `String[]` parts. A record builds `equals`, `hashCode` and `toString` from
its parts, and for an array part those compare the array object, not the lines inside it. That is
why this file needs 30 lines of hand-written methods and a 92-line test. The getters also hand out
the real arrays, so a caller can change the object after it was built.

### Suggested fix

Use `List<String>` and copy the lines in the constructor. Then the three hand-written methods and
most of `PythonSourceTest` can go:

```java
public record PythonSource(List<String> globalCode, List<String> code, List<String> expression) {
    public PythonSource(List<String> globalCode, List<String> code, List<String> expression) {
        this.globalCode = copyOrEmpty(globalCode);
        this.code = copyOrEmpty(code);
        this.expression = copyOrEmpty(expression);
    }
}
```

Write the constructor out in full, not as a compact one, so no parameter is reassigned. Treat all
three parts the same, so no getter returns null; `of(String, String, String)` already turns null
into an empty array today, so behaviour does not change, and
`PythonSourceTest.equalsDistinguishesNullFromEmptyExpression` can be deleted. `FunctionDefinition`
keeps its `String[]` fields, so add `toArray(String[]::new)` in the three places that set them.
Two small things in the same file: move the class comment above `public record` (it now sits above
`import java.util.Arrays;`, so it describes the import), and delete
`of(String[], String[], String[])` at line 43, which only repeats the constructor.

Not for this PR: `definition/parser/FunctionDefinitionParser.java:76` checks
`rawFunction.expression() != null`, and the parse path never gives null, on `main` either. So that
part of the check is always true. Old issue, please leave it as it is here.

KasparMetsa (changes requested):

ksml/src/main/java/io/axual/ksml/operation/parser/OperationParser.java
```diff
@@ -31,29 +34,48 @@
...
-import io.axual.ksml.parser.TopologyResourceAwareParser;
+import io.axual.ksml.parser.TopologyResourceFields;
 import io.axual.ksml.parser.TopologyResourceParser;
 import io.axual.ksml.store.StoreType;
 import lombok.Getter;
 
 import java.util.List;
+import java.util.function.BiFunction;
 
 @Getter
```
KasparMetsa: ### Current problem

`@Getter` on the class covers every non-static field, so the new `resourceFields` field gets a
public `getResourceFields()`. The point of holding a helper object instead of extending a parent
class was to stop handing the helper methods to subclasses. A public getter hands the whole helper
to every caller instead. The class already has `resources()`, `functionField()`, `topicField()` and
`lookupField()` as `protected` methods, so nothing needs the field itself.

### Suggested fix

Add `@Getter(AccessLevel.NONE)` on the field, or drop `@Getter` from the class and put it on
`defaultShortName`, `defaultLongName` and `type` one by one. I found no caller of
`getResourceFields()` in the PR.

ksml/src/main/java/io/axual/ksml/parser/TopologyResourceFields.java
```diff
@@ -0,0 +1,66 @@
...
+// Field-builder helpers available once the full resource set is known, i.e. once all
+// streams/tables/topics have been parsed and assembled into a TopologyResources (which is a
+// TopologyBaseResources). This is the phase pipelines and operations run in, so topics can be
+// referenced by name (topicField) in addition to the base-phase capabilities (delegated to a
+// composed TopologyBaseResourceFields, so the base-phase logic isn't duplicated).
+public class TopologyResourceFields {
+    private final TopologyBaseResourceFields base;
+    private final TopologyResources resources;
+
+    public TopologyResourceFields(TopologyResources resources) {
```
KasparMetsa, line 41: ### Current problem

`openspec/changes/archive/2026-07-23-simplify-baseparser-composition/proposal.md:5` calls it a bug
that `TopologyBaseResourceAwareParser` and `TopologyResourceAwareParser` each held the same object,
only to see it as a narrower type. The new code does the same: `base` and `resources` here point to
one object, and both classes have a `resources()` with the same null check. I do not think this can
be removed, because `TopologyBaseResourceFields` is also used on its own and needs the wider type,
and a cast would be worse. So this is only about the proposal text and about the check.

Second point: the `TopologyException` in `resources()` can never be shown. `functionField` and
`topicField` read the field directly, for example `(name, context) -> resources.topic(name)`, so a
null field gives a `NullPointerException` while parsing. This is the same on `main`, so it is not
new, but this PR rewrites these lines.

### Suggested fix

Say in the proposal that this one part was kept on purpose. Then put
`Objects.requireNonNull(resources, "...")` in both constructors and let `resources()` just return
the field, so the check happens once, early, and with a message that a reader can act on.

ksml/src/main/java/io/axual/ksml/parser/DefinitionParser.java
```diff
@@ -20,357 +20,22 @@
...
 
-import java.time.Duration;
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
-
-import static io.axual.ksml.data.schema.DataSchemaConstants.NO_TAG;
 
 public abstract class DefinitionParser<T> extends BaseParser<T> implements StructsParser<T> {
```
KasparMetsa: ### Current problem

The change is called "simplify-baseparser-composition", but `DefinitionParser<T>` still extends
`BaseParser<T>`. `BaseParser` only holds three `protected` helpers, `parseBoolean`, `parseInteger`
and `parseString`. That is the same "hand helper methods down through a parent class" idea this PR
removes everywhere else, and no class under `DefinitionParser` uses them: the callers are the
classes in `io.axual.ksml.schema.parser` and `definition/parser/JoinTargetDefinitionParser.java`,
and all of those extend `BaseParser` directly.

### Suggested fix

Write `public abstract class DefinitionParser<T> implements StructsParser<T>`. `StructsParser`
already reaches `Parser<T>` through `ParserWithSchemas`, and `BaseParser<T>` is never used as a
variable or parameter type, so about 90 classes lose a parent they do not need. If some class does
not compile after that, please write it in the proposal, so the folder name does not promise
something the PR does not do.

ksml/src/test/java/io/axual/ksml/parser/StructsParserLazyTest.java
```diff
@@ -0,0 +1,79 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+// Covers the memoize-once and fatal-error-wrapping contracts StructsParser.lazy()
+// must preserve from DefinitionParser's historic parse()/schemas() implementation.
+class StructsParserLazyTest {
```
KasparMetsa, line 36: ### Current problem

`lazy_wrapsParseExceptionsAsFatalError` passes whether `lazy()` wraps the error or not.
`IllegalStateException` is already a `RuntimeException`, and its message is already `"boom"`, so
both checks hold for the original exception. Remove the try/catch from `StructsParser.lazy` and the
test still passes.

### Suggested fix

`FatalError.report` returns `new RuntimeException(t.getMessage())` and drops the cause, so check
the exact class and the missing cause:

```java
assertThatThrownBy(() -> lazy.parse(node))
        .isExactlyInstanceOf(RuntimeException.class)
        .hasMessage("boom")
        .hasNoCause();
```

This also answers the earlier AssertJ question in this file: on `main`, 121 of 166 test classes
under `ksml/src/test/java` import `org.assertj`, so AssertJ is the normal style here. Last small
thing: `lazy_buildsDelegateAtMostOnce` asserts exactly 1, so `buildsDelegateExactlyOnce` is a
better name.

KasparMetsa (changes requested):

developerdocs/parser-architecture.md
```diff
@@ -0,0 +1,338 @@
...
+
+`StructsParser<T>` — the interface almost everything in `DefinitionParser` is built around —
+`extends ParserWithSchemas<T>` and narrows `schemas()`'s return type to
+`List<StructSchema>` specifically (a legal covariant override, since `List<StructSchema>` is a
+subtype of `List<? extends DataSchema>`), and adds a default `schema()` that returns the single
+schema directly if there's only one, or wraps several in a `UnionSchema` otherwise.
+
+## 4. `DataSchema` — what the shape actually *is*
+
+`DataSchema` is a small sealed family of value types describing a shape, independent of any
```
KasparMetsa: ### Current problem

Line 102 calls `DataSchema` "a small sealed family of value types". It is not sealed.
`ksml-data/src/main/java/io/axual/ksml/data/schema/DataSchema.java:43` reads
`public class DataSchema implements DataEquals`, so any class can extend it.

### Suggested fix

Drop the word "sealed", or write that the set is closed by agreement, not by the compiler.

ksml/src/main/java/io/axual/ksml/definition/FunctionDefinition.java
```diff
@@ -111,14 +103,14 @@ protected TopologyException functionResultError(String message) {
         throw new TopologyException(message + ": function=" + name() + ", type=" + type() + ", resultType=" + resultType());
     }
 
-    private FunctionDefinition(String type, String name, ParameterDefinition[] parameters, String[] globalCode, String[] code, String[] expression, UserType resultType, List<String> storeNames) {
+    private FunctionDefinition(String type, String name, ParameterDefinition[] parameters, PythonSource source, UserType resultType, List<String> storeNames) {
```
KasparMetsa, line 107: ### Current problem

The constructor reads `source.expression()`, `source.code()` and `source.globalCode()` without a
null check, and the public `as(...)` methods accept null. Before this PR,
`as(type, name, params, (String) null, (String) null, null, ...)` was a normal call and gave empty
arrays; `ToOperationParsersTest` on `main` calls it that way. Now
`as(type, name, params, null, resultType, stores)` throws `NullPointerException` with no message.

### Suggested fix

Add `Objects.requireNonNull(source, "PythonSource is required")` at the top of the constructor, so
a wrong call fails with a clear message.

ksml/src/main/java/io/axual/ksml/definition/PythonSource.java
tonvanbart: Changed the constructor and call sites in FunctionDefinition. I left the test as is, IMO it's not in the way and there if we decide on another refactoring in the future.

ksml/src/main/java/io/axual/ksml/parser/DefinitionParser.java
tonvanbart: Tried the suggested change and re-ran a full build which still passes.

ksml/src/test/java/io/axual/ksml/parser/StructsParserLazyTest.java
tonvanbart: Fixed as per the suggestion, I will do a separate commit to change everything to AssertJ.

developerdocs/parser-architecture.md
tonvanbart: Removed "sealed"

ksml/src/main/java/io/axual/ksml/parser/TopologyResourceFields.java
tonvanbart: Added the proposed fix

ksml/src/main/java/io/axual/ksml/definition/FunctionDefinition.java
tonvanbart: Done

developerdocs/parser-architecture.md
tonvanbart: Fixed

KasparMetsa (approved):

PR #698, opened 2026-08-31
Fix the documented bugs and dead code

tonvanbart:
This PR fixes most of the findings from the bugs-and-dead-code-dandidates.md document. Findings under the heading "lower confidence" have been left alone as some of these are still open for discussion.


ksml-runner/src/main/java/io/axual/ksml/runner/KSMLRunner.java
```diff
@@ -754,11 +757,16 @@ static KSMLRunnerConfig readConfiguration(File configFile) {
...
 
-    static ErrorHandler getErrorHandler(ErrorHandlingConfig.ErrorTypeHandlingConfig config) {
+    static ErrorHandler getErrorHandler(ErrorHandlingConfig.ErrorTypeHandlingConfig config, boolean retrySupported) {
         final var handlerType = switch (config.handler()) {
             case CONTINUE -> ErrorHandler.HandlerType.CONTINUE_ON_FAIL;
             case STOP -> ErrorHandler.HandlerType.STOP_ON_FAIL;
-            case RETRY -> ErrorHandler.HandlerType.RETRY_ON_FAIL;
+            case RETRY -> {
+                if (!retrySupported) {
+                    throw new ConfigException("retryOnFail is not a supported error handler here; only stopOnFail or continueOnFail are allowed");
```
KasparMetsa: Hey this comment was generated by Claude but I checked and it makes sense to me to fix docs and schema regarding this:

### 1. Docs and schema still promise `retryOnFail` for consume and process, but the code now rejects it

Code in PR, `ksml-runner/src/main/java/io/axual/ksml/runner/KSMLRunner.java`:

```java
static void setupErrorHandling(ErrorHandlingConfig errorHandling) {
    if (errorHandling != null) {
        // Only the produce (serialization) path has a Kafka Streams RETRY response; consume
        // (deserialization) and process handlers only support CONTINUE/STOP and must reject
        // retryOnFail here rather than let it crash the Streams thread on first use.
        ExecutionContext.INSTANCE.errorHandling().setConsumeHandler(getErrorHandler(errorHandling.consumerErrorHandlingConfig(), false));
        ExecutionContext.INSTANCE.errorHandling().setProduceHandler(getErrorHandler(errorHandling.producerErrorHandlingConfig(), true));
        ExecutionContext.INSTANCE.errorHandling().setProcessHandler(getErrorHandler(errorHandling.processErrorHandlingConfig(), false));
    }
}

static ErrorHandler getErrorHandler(ErrorHandlingConfig.ErrorTypeHandlingConfig config, boolean retrySupported) {
    ...
    case RETRY -> {
        if (!retrySupported) {
            throw new ConfigException("retryOnFail is not a supported error handler here; only stopOnFail or continueOnFail are allowed");
        }
        yield ErrorHandler.HandlerType.RETRY_ON_FAIL;
    }
```

This is a correct fix for a real bug (before this PR, `retryOnFail` on `consume`/`process` passed
validation and then crashed the Kafka Streams thread the first time an error actually happened).
But three places that tell the user what values are valid were not updated to match:

- `docs/reference/configuration-reference.md:168`: "Each error type (`consume`, `process`,
  `produce`) supports these properties: ... `handler` ... (`continueOnFail`, `stopOnFail`,
  `retryOnFail`)"
- `docs/getting-started/schema-validation.md:81` and `:347`: lists `retryOnFail` as a valid,
  suggested value for `errorHandling.consume.handler` with no restriction mentioned
- `ksml-runner/src/main/java/io/axual/ksml/runner/config/ErrorHandlingConfig.java:107`, the
  `@JsonPropertyDescription` on the shared `handler` field used by consume/process/produce alike:
  "Controls the behaviour of handling the error, which can be Stop KSML (stopOnFail), Continue KSML
  (continueOnFail) and Retry (retryOnFail)"

That last one is not just a comment: it feeds the generated JSON schema. I ran the schema
generator and confirmed `docs/ksml-runner-spec.json` still lists
`"enum" : [ "stopOnFail", "continueOnFail", "retryOnFail" ]` as valid for all three handler types.
So an editor with schema validation on will suggest `retryOnFail` for `consume:` and `process:`,
and the docs page that explains this exact schema validation feature uses `consume.handler:
retryOnFail` as its own example of a valid suggestion. A user who follows the docs will get a
startup `ConfigException`, not the behaviour promised.

Suggested change: update the three text spots above to say `retryOnFail` is produce-only, for
example:

```
| `handler`    | String  | stopOnFail     | Error handling strategy (`continueOnFail`, `stopOnFail`, and for `produce` only, `retryOnFail`) |
```

ksml-runner/src/test/java/io/axual/ksml/runner/KSMLRunnerErrorHandlerTest.java
```diff
@@ -23,33 +23,55 @@
...
 
-    @ParameterizedTest(name = "handler {0} -> {1}")
+    @ParameterizedTest(name = "handler {0} -> {1}, retry-capable")
     @CsvSource({
             "STOP,     STOP_ON_FAIL",
             "CONTINUE, CONTINUE_ON_FAIL",
             "RETRY,    RETRY_ON_FAIL"
     })
-    void mapsEveryHandlerToItsHandlerType(Handler handler, ErrorHandler.HandlerType expected) {
+    void mapsEveryHandlerToItsHandlerTypeWhenRetryIsSupported(Handler handler, ErrorHandler.HandlerType expected) {
```
KasparMetsa, line 44: Code in PR, `ksml-runner/src/test/java/io/axual/ksml/runner/KSMLRunnerErrorHandlerTest.java`:

```java
void mapsEveryHandlerToItsHandlerTypeWhenRetryIsSupported(Handler handler, ErrorHandler.HandlerType expected) {
    ...
    assertThat(KSMLRunner.getErrorHandler(cfg, true).handlerType()).isEqualTo(expected);
}
...
void rejectsRetryOnFailWhenNotSupported() {
    ...
    assertThatThrownBy(() -> KSMLRunner.getErrorHandler(cfg, false))
```

All three new/changed tests call `getErrorHandler(config, boolean)` directly with a hand-typed
`true`/`false`. None of them go through `setupErrorHandling`, which is the one place that decides
which of the three real channels gets `true` and which get `false`. The only test that calls
`setupErrorHandling` is `KSMLRunnerHelpersTest.setupErrorHandlingWiresAllChannels`, and it only
checks the logger names on a default config that never turns on `RETRY`:

```java
void setupErrorHandlingWiresAllChannels() throws Exception {
    KSMLRunner.setupErrorHandling(new ErrorHandlingConfig());
    assertThat(reflectHandler("consumeHandler").loggerName()).isEqualTo("ConsumeError");
    assertThat(reflectHandler("produceHandler").loggerName()).isEqualTo("ProduceError");
    assertThat(reflectHandler("processHandler").loggerName()).isEqualTo("ProcessError");
}
```

Suggested change: add a test that builds an `ErrorHandlingConfig` with `consume.handler:
retryOnFail` (and one with `process.handler: retryOnFail`), calls `setupErrorHandling` with it, and
asserts it throws `ConfigException`. Also add one with `produce.handler: retryOnFail` that asserts
it does not throw.

ksml-runner/src/main/java/io/axual/ksml/runner/KSMLRunner.java
```diff
@@ -547,9 +547,12 @@ static Map<String, String> mergeNotationConfigs(NotationConfig notationConfig, M
      */
     static void setupErrorHandling(ErrorHandlingConfig errorHandling) {
         if (errorHandling != null) {
-            ExecutionContext.INSTANCE.errorHandling().setConsumeHandler(getErrorHandler(errorHandling.consumerErrorHandlingConfig()));
-            ExecutionContext.INSTANCE.errorHandling().setProduceHandler(getErrorHandler(errorHandling.producerErrorHandlingConfig()));
-            ExecutionContext.INSTANCE.errorHandling().setProcessHandler(getErrorHandler(errorHandling.processErrorHandlingConfig()));
+            // Only the produce (serialization) path has a Kafka Streams RETRY response; consume
+            // (deserialization) and process handlers only support CONTINUE/STOP and must reject
+            // retryOnFail here rather than let it crash the Streams thread on first use.
+            ExecutionContext.INSTANCE.errorHandling().setConsumeHandler(getErrorHandler(errorHandling.consumerErrorHandlingConfig(), false));
```
KasparMetsa: This is a small thing, please feel free to ignore if you disagree with this point:

Code in PR, `ksml-runner/src/main/java/io/axual/ksml/runner/KSMLRunner.java:553-555`:

```java
ExecutionContext.INSTANCE.errorHandling().setConsumeHandler(getErrorHandler(errorHandling.consumerErrorHandlingConfig(), false));
ExecutionContext.INSTANCE.errorHandling().setProduceHandler(getErrorHandler(errorHandling.producerErrorHandlingConfig(), true));
ExecutionContext.INSTANCE.errorHandling().setProcessHandler(getErrorHandler(errorHandling.processErrorHandlingConfig(), false));
```

Nothing in the type system says "produce may retry, consume and process may not" - that fact only
lives in these three plain `true`/`false` literals, typed by hand, in this one place. A future
fourth caller, or someone re-ordering these three lines, can pass the wrong value with nothing to
stop them. The `ConfigException` message that gets thrown also does not say which channel
(consume/produce/process) rejected `retryOnFail`, because the method has no way to know.

Suggested change: let the caller pass a name/channel identifier instead of a bare boolean, or move
the "which handler types are supported here" knowledge onto the channel itself (for example, a
small `enum ErrorChannel { CONSUME(false), PRODUCE(true), PROCESS(false); }`) so a new call site
gets the right answer by construction instead of by a hand-typed literal.

developerdocs/bugs-and-dead-code-candidates.md
KasparMetsa, line 138: Code in PR, `developerdocs/bugs-and-dead-code-candidates.md`, section 5:

```
### 5. `CsvDataObjectConverter`'s structured-CSV branch is dead due to a type-mismatched `equals`
...
if (value instanceof DataList valueList && DEFAULT_TYPE.equals(valueList.type())) {
...
No test exercises this class.
```

Section 6:

```
### 6. `JoinTargetDefinitionParser` rejects inline `table:`/`globalTable:` join targets **(verified directly)**
...
if (parseString(node, Operations.Join.WITH_TABLE) != null) {
...
No test exercises an inline `table:`/`globalTable:` join target
```

And in the "Lower confidence" list:

```
- `CachedPatternResolver` (ksml-kafka-clients) configures its Guava caches with
  `expireAfterAccess`/`expireAfterWrite(Duration.ZERO)`, which per Guava semantics makes caching
  a no-op.
```

All three describe the code as it was *before* this PR. This same diff changes
`CsvDataObjectConverter.java`'s `equals` check to `isAssignableFrom` and adds
`CsvDataObjectConverterTest.java`; changes `JoinTargetDefinitionParser.java`'s `parseString(...)` to
`node.get(...) != null` and adds `KSMLJoinInlineTargetTest.java`; and changes
`CachedPatternResolver.java`'s `Duration.ZERO` to `Duration.of(365, ChronoUnit.DAYS)`. Every other
fixed item in this same file was struck through and marked "**(Fixed ...)**" or "**(Already
resolved)**" - these three were not, so they now read as still-open next to items that really are
still open (like the "namespace parameter" and "fileDescriptor" bullets right next to the stale
`CachedPatternResolver` one).

Suggested change: strike through and mark these three the same way the rest of the file already
does, for example:

```
- ~~`CachedPatternResolver` ... `Duration.ZERO`, which ... makes caching a no-op.~~ **(Fixed)** -
  changed to a 365-day expiry.
``

KasparMetsa (changes requested):

ksml-runner/src/main/java/io/axual/ksml/runner/KSMLRunner.java
tonvanbart: Fixed docs

ksml-runner/src/test/java/io/axual/ksml/runner/KSMLRunnerErrorHandlerTest.java
tonvanbart: Stored the combination of consume/produce/process and retrySupported true/false in an enum; added test

ksml-runner/src/main/java/io/axual/ksml/runner/KSMLRunner.java
tonvanbart: Added enum

developerdocs/bugs-and-dead-code-candidates.md
tonvanbart: Revised the document

KasparMetsa (approved):

PR #702, opened 2026-09-03
Fix small code issues

tonvanbart:
Fix some of the Sonar findings: S1168 + S1172 + S2589 + S112


ksml-test-runner/src/main/java/io/axual/ksml/testrunner/FieldExtractor.java
```diff
@@ -88,40 +89,40 @@ JsonNode requireArray(String field) {
...
-        return YAML_MAPPER.convertValue(child, LinkedHashMap.class);
+        return Optional.of(YAML_MAPPER.convertValue(child, LinkedHashMap.class));
     }
 
     /**
-     * Get an optional array field as a list of strings, returning null if absent.
+     * Get an optional array field as a list of strings, returning empty if absent.
      */
-    List<String> optionalStringList(String field) {
+    Optional<List<String>> optionalStringList(String field) {
```
KasparMetsa: ## 1. The Optional fix is undone right after it is made

**File:** `ksml-test-runner/src/main/java/io/axual/ksml/testrunner/FieldExtractor.java` (lines 93-125) and `ksml-test-runner/src/main/java/io/axual/ksml/testrunner/TestDefinitionParser.java` (line 301), `ksml-test-runner/src/main/java/io/axual/ksml/testrunner/ProduceBlock.java` (line 65)

**Code in PR:**
```java
// FieldExtractor.java
Optional<List<String>> optionalStringList(String field) { ... }
Optional<Map<String, Object>> optionalMap(String field) { ... }
Optional<Long> optionalLong(String field) { ... }
```
```java
// TestDefinitionParser.java, parseAssertBlocks()
var stores = f.optionalStringList(KSMLTestDSL.Assert.STORES).orElse(null);
```
```java
// ProduceBlock.java, of()
var block = new ProduceBlock(to, messages.orElse(null), generator.orElse(null), count.orElse(null));
```

**Suggested change:** For the list and map methods, return an empty collection directly (`List.of()`, `Map.of()`) instead of wrapping in `Optional`. Keep `AssertBlock.stores`, `ProduceBlock.messages`, and `ProduceBlock.generator` as plain, never-null fields. Drop the `.orElse(null)` calls.

**Reason:** The Sonar rule this PR fixes (S1168) asks for empty collections instead of null. The PR changes the return type to `Optional`, but then the very next step in every caller turns it right back into `null` with `.orElse(null)`. The record fields (`AssertBlock.stores`, `ProduceBlock.messages`, `ProduceBlock.generator`) are still nullable and still null-checked further down (see `AssertBlock.validate()` and `ProduceBlock.validate()`). So the null is not removed. It is only moved one line down. `Optional<Long>` for a single value is a reasonable fix, but for the list/map methods the round trip through `Optional` adds code without solving the problem.

KasparMetsa (changes requested):

ksml-test-runner/src/main/java/io/axual/ksml/testrunner/FieldExtractor.java
tonvanbart: Made the changes and adjusted the tests, also added testcases for invalid produce blocks.

KasparMetsa:
**CRITICAL: merge conflict brings back the ForeignDict bug risk**

GitHub shows this PR as "CONFLICTING" with `main`. The conflict is in `AssertionRunner.java`, in the exact method this PR touches.

Code in this PR's branch:

```java
if (stream == null) {
    return TestResult.error(suiteName, testName,
            "Assert block references undeclared stream '" + block.on() + "'");
}
var records = recordCache.computeIfAbsent(stream.topic(), _ -> collectOutputRecords(stream));
args.add(Pair.of("records", ProxyUtil.toPython(records)));
```

Code on current `main`:

```java
var records = recordCache.computeIfAbsent(stream.topic(), k -> collectOutputRecords(stream));
args.add(Pair.of("records", PYTHON_MAPPER.toRealPythonValue(pythonContext.context(), records)));
```

`ProxyUtil.toPython(records)` is the old, buggy way. It wraps `records` in a proxy object, not a real Python list. That is the exact bug fixed in #707/#708 (`copy.deepcopy()` crashing with `TypeError: cannot pickle 'ForeignDict' object`). `PYTHON_MAPPER.toRealPythonValue(...)` is the fixed way.

This branch was made before that fix landed on `main`. So the two sides disagree:
- This PR's side: has a null check for `stream`, but uses the old, buggy `ProxyUtil.toPython`.
- `main`'s side: uses the fixed `PYTHON_MAPPER.toRealPythonValue`, but is missing the null check.

If this conflict gets resolved by picking this PR's side (or without looking closely), the ForeignDict bug comes back in test assertion code.

Suggested fix: rebase on current `main`, then resolve by hand, keeping both `PYTHON_MAPPER.toRealPythonValue(...)` and the `if (stream == null)` check.

KasparMetsa:
**HIGH: an existing test was made stricter for files it did not add**

`KSMLTestRunnerTest.java`:

```java
// before
@ValueSource(strings = {
        "duplicate-stream-topic.yaml",
        "undefined-stream-reference.yaml",
        "bare-vendor-avro.yaml",
        "generator-bad-field-type-test.yaml"
})
void invalidDefinitionsReturnNonPass(String testFile) {
    ...
    assertTrue(results.stream().anyMatch(r -> r.status() != TestResult.Status.PASS), ...);
}

// after
@ValueSource(strings = {
        "duplicate-stream-topic.yaml",
        "undefined-stream-reference.yaml",
        "bare-vendor-avro.yaml",
        "generator-bad-field-type-test.yaml",
        "produce-block-neither-messages-nor-generator.yaml",
        "produce-block-both-messages-and-generator.yaml"
})
void invalidDefinitionsReturnError(String testFile) {
    ...
    assertTrue(results.stream().anyMatch(r -> r.status() == TestResult.Status.ERROR), ...);
}
```

This PR needs to check that its two new invalid YAML files return `ERROR`. That part makes sense. But instead of adding a new, separate test for just those two files, the PR renamed the old test and tightened its check for the four files that were already there. Those four files used to only need to be "not PASS" (so `FAIL` was fine too). Now they must be exactly `ERROR`.

Why this matters: if a future change makes one of those four files return `FAIL` instead of `ERROR` (which may be a correct, wanted outcome), this test breaks, even though nothing is actually wrong. The test now checks something it was never written to check, for files this PR has nothing to do with.

Suggested fix: keep `invalidDefinitionsReturnNonPass` as it is, for the four old files. Add a new, separate parameterized test (for example `newInvalidProduceBlocksReturnError`) for just the two new files, with the `== ERROR` check.

KasparMetsa:
**HIGH: `ProduceBlock.of()` is not needed, and makes two similar classes inconsistent**

`ProduceBlock.java` adds a new static method:

```java
public static ProduceBlock of(String to, List<TestMessage> messages,
                               Map<String, Object> generator, Optional<Long> count) {
    var block = new ProduceBlock(to, messages, generator, count.orElse(null));
    block.validate();
    return block;
}
```

`TestDefinitionParser.java` then does:

```java
var block = ProduceBlock.of(to, messages, generator, count);
```

But the sibling class `AssertBlock` is still built the old, plain way, right next to this code in the same file:

```java
var block = new AssertBlock(on, stores, code);
block.validate();
```

Now the file has two different ways to build a very similar "block" object: one hides `.validate()` inside a factory method, the other calls the constructor and `.validate()` separately. A reader now has to remember which block type uses which pattern.

The only real reason for `.of()` is to turn `Optional<Long> count` into a plain `Long`. That does not need a new method. It can be done directly at the call site, matching the pattern already used before this PR:

```java
var count = f.optionalLong(KSMLTestDSL.Produce.COUNT).orElse(null);
var block = new ProduceBlock(to, messages, generator, count);
block.validate();
```

Worth noting: `ProduceBlock.of()` is called in exactly one place. Every test file for `ProduceBlock` still builds it with `new ProduceBlock(...)` directly and skips `.of()` completely.

Suggested fix: remove `ProduceBlock.of()` and unwrap `Optional<Long>` at the call site instead, matching how `AssertBlock` is built.

KasparMetsa:
**MEDIUM: `FieldExtractor`'s three "optional..." methods no longer agree with each other**

```java
Optional<Long> optionalLong(String field) { ... }      // wraps result in Optional
Map<String, Object> optionalMap(String field) { ... }  // returns Map.of() directly, no Optional
List<String> optionalStringList(String field) { ... }  // returns List.of() directly, no Optional
```

All three method names start with "optional", but only one of them actually returns `Optional<...>`. The other two just return an empty collection when the field is missing. A reader cannot tell from the method name alone what kind of "optional" this is, which makes it easy to misuse one by accident (for example expecting `.isPresent()` on a method that never returns `Optional`).

Suggested fix: pick one style for all three. Either keep `optionalLong` as `Optional<Long>` (reasonable here, since `0` is a real, meaningful count value, so it cannot double as "missing"), and rename the other two so their name does not say "optional" anymore (for example `mapField`/`stringListField`), matching what they actually return.

tonvanbart:
Re: "CRITICAL: merge conflict brings back the ForeignDict bug risk": fixed as per the suggestion

Re: "HIGH: an existing test was made stricter for files it did not add": This is intentional; any malformed yaml defintions should return `ERROR`, not just "not `PASS`". I realized this when I added the new cases so renamed the test and tightened the assertion.

Re: "HIGH: ProduceBlock.of() is not needed, and makes two similar classes inconsistent": fixed as per the suggestion

Re: "MEDIUM: FieldExtractor's three "optional..." methods no longer agree with each other": fixed as per the suggestion

KasparMetsa (approved):

PR #708, opened 2026-09-11
Fix: Python receives proxy objects instead of real dict/list (backport of #707 for 1.3.1)

KasparMetsa:
Backport of #707 to the 1.3 line, for a 1.3.1 patch release.

Same bug and fix as #707: message keys, values, and state-store aggregated values were wrapped in a Java proxy object before being handed to Python. It looks like a dict on the surface, but `copy.deepcopy()` crashes with `TypeError: cannot pickle 'ForeignDict' object`, and `dict(value)` only makes a shallow copy. This broke real pipelines that were upgraded to 1.3.0.

A few differences from #707, all needed only because this branch predates later, unrelated `main` changes:
* One new test (`PythonNativeConversionTest`) used a `PythonSource` helper class that does not exist on this branch yet; adjusted it to call `FunctionDefinition.as(...)` the way this branch's code already does.
* Two test files needed an added `import static org.assertj.core.api.Assertions.assertThat[ThrownBy]` - those tests already relied on AssertJ fluent assertions on `main`, but the import itself was added by unrelated `main` work after 1.3.0, so it was missing here.
* Six proxy-store unit tests (`ProxyUtilTest`, `KeyValueStoreProxyTest`, etc.) were split out of one bigger file on `main` in July, before this fix even started - unrelated to this bug on its own. But #707 rewrote the assertions in exactly these files to check the new dict/list-returning behavior, so they're included here too: leaving them out meant the changed `ProxyUtil` code had no dedicated test on this branch, and new-code coverage fell under the quality gate's minimum.
* `pom.xml`'s surefire/failsafe coverage wiring was broken on this branch: Jacoco's `prepare-agent` goals had no `propertyName` set, so they wrote the coverage agent flag to a property nobody read, and `argLine`/`failsafeArgLine` used early-bound `${...}` placeholders instead of the late-bound `@{...}` ones Jacoco needs. Net effect: the coverage agent never attached to any test on this branch, so Sonar's new-code coverage check saw almost nothing even though the tests already covered this fix's changed lines. This is the same fix `main` already has in commit `93cd24b9` ("Fix Jacoco coverage") - this branch just forked before that fix landed.

Also fixed: the `docs/release-notes.md` entry for this fix had landed under the 1.1.0 section on `main` (a genuine mistake in #707, not something specific to this backport) - moved it here to a new 1.3.1 (unreleased) section. Worth fixing the same way on `main`.

Tested locally: full test suite (1614 tests in `ksml` module, 1576+ across the whole reactor) green under a real GraalVM JDK, including `VulnerabilitiesTest` (the security-exploit regression suite), the six proxy-store tests, and the new `PythonNativeConversionTest`/`testDeepcopyOnStateStoreRead` tests that check the fix directly. Also verified against the actual reported production bug: reproduced the exact `TypeError: cannot pickle 'ForeignDict' object` crash from `add_date_to_value`'s `deepcopy()` call using the real `axual/ksml:1.3.0` image and an unmodified production KSML definition, then confirmed a locally-built image from this branch fixes it with no changes to that definition.


pantelis17:
commit 35ae435d ("Fix Jacoco coverage wiring so Sonar can see real numbers") changes pom.xml's surefire/failsafe argLine wiring. It's correct and matches main, but it isn't called out in the PR description alongside the other three disclosed differences. Can you add a line about it there for completeness?

pantelis17 (approved):
Other than the PR description looks good, approved

KasparMetsa:
Good catch, added a line about it in the description.

